PR #123, opened 2015-12-07
Additional search path for lua.

bentley:
OpenBSD installs the headers under /usr/local/include/lua-5.3. This change is enough to pick it up.


SuperV1234:
Thanks!

PR #129, opened 2016-02-08
Lua functions for changing style, music and rotation

MultHub:
I was so close to giving up.
Have fun with this thing.

For l_setLevel: `l_setLevel("Packs/<pack>/<level id>")` was the part that almost made me give up

Short summary of new Lua functions:
`l_setLevel(string mId)`
`u_setMusic(string mId)`
`s_setStyle(string mId)`
`l_setRotation(float mValue)`, see commit 1cef3a741933c36b66b7605ac591ece1c8fe2076
`l_getRotation()`


SuperV1234:
Thanks for the pull request. I will merge and then (temporarily) remove `l_setLevel` because I need to make sure it works properly with high scores and profiles.

PR #132, opened 2016-03-08
Segment time: use float

MultHub:
It's something. Something useful.
Because there are some people who care about perfect timing.

Random question: any useful build instructions for Windows? The instructions in the README look a bit outdated, looks like MinGW doesn't have GCC 5.x


SuperV1234:
Looks good. Thanks!

PR #137, opened 2016-05-04
boost checks only for MinGW, link against pthread on Unix

darealshinji:
If I understand it right, boost will only be required for MinGW builds (at least it isn't used on Linux): https://github.com/SuperV1234/SSVOpenHexagon/issues/130#issuecomment-216218123


SuperV1234:
There is a choice to be made here - I would like your input as well:

1) Use [mingw-std-threads](https://github.com/meganz/mingw-std-threads/), that provides `std::thread`, and use boost only to provide `boost::future`. The license does not look very promising though.

2) Use boost both for threads and futures.

darealshinji:
I'd pick the one which has better performance and/or is more portable. As for the license, I don't think there should be any problems. mingw-std-threads is released under a 2-clause BSD license, so you could even use it in proprietary closed-source projects.

SuperV1234:
I've pushed three commits that make use of mingw-std-threads and boost-future (02a7dd0ce618f47f1c1acf007b3c84177fac8757, 3c71dda729931ae90916c91764830985692e078d, e3da34ed15ca77faa3b592062cb6f12647d1c3d9).

mingw-std-threads is now a submodule of SSVOpenHexagon. I've tried cloning and building from scratch with MinGW from nuwen.net and it works properly (note that I also re-built SFML, Lua and ZLib).

PR #163, opened 2020-06-18
Added Lua Mouse Support (u_isButtonPressed)

Morxemplum:
This is a feature pack developers have wanted for a long time, now you can track the inputs of the mouse alongside the keyboard in Lua code, removing "[Keyboard Required]" from (most) levels altogether.


Morxemplum:
I decided to rename the function so it's more trivial.

PR #167, opened 2020-06-18
Fix Scaling Bug On Message Text

Morxemplum:
This one line of code fixes the bug where the message text doesn't scale accordingly to the resolution of the window, leading to the text being huge on smaller resolution screens, and being smaller on higher resolutions.


SuperV1234:
Thanks1

PR #170, opened 2020-06-19
Lua camera set angle

Zly-u:
Moved Lua Context lower in order to fix some possible and existed conflict with Lua execution in levels, for example with l_setRotation.


SuperV1234:
Will manually merge to avoid conflicts and unrelated changes.

PR #171, opened 2020-06-19
added `m_messageAddImportantSilent` lua function

Zly-u:
Implemented a flag for `m_messageAddImportant` to toggle the sound on message appear.
Basically for text updating purposes that people usually use it for

### DIFF
--- a/src/SSVOpenHexagon/Core/HGScripting.cpp
+++ b/src/SSVOpenHexagon/Core/HGScripting.cpp
@@ -67,11 +67,20 @@ void HexagonGame::initLua_Messages()
     });
 
     lua.writeVariable(
-        "m_messageAddImportant", [this](string mMsg, float mDuration, bool mSoundTog) {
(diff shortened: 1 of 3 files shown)


src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -61,17 +61,17 @@ void HexagonGame::initLua_Messages()
...
     });
 
     lua.writeVariable(
-        "m_messageAddImportant", [this](string mMsg, float mDuration) {
+        "m_messageAddImportant", [this](string mMsg, float mDuration, bool mSoundTog) {
             eventTimeline.append<Do>([=, this] {
                 if(Config::getShowMessages())
                 {
-                    addMessage(mMsg, mDuration);
+                    addMessage(mMsg, mDuration, mSoundTog);
```
Morxemplum, line 74: Does this really have to be mandatory? This easily can break compatibility.

PR #172, opened 2020-06-19
Make functions previously exclusive in JSON events into Lua functions

Morxemplum:
This is an addition of several new Lua functions into the Lua library. A lot of these functions were based off event functions that were exclusive to JSON in 1.92, which means that anybody who wants to port their packs from 1.92 can do this more easily if their levels were very reliant on events. 

I didn't do menu, because that can be substituted for "kill" and clearTimeline has been completely replaced with clearWalls, which provides much more useful functionality of clearing out all of the walls in a current level (because clearTimeline was pointless). 

Sorry for the merge issues. This probably has to do with the fact I have to add all the new Lua functions to the whitelist so the debugger doesn't make pointless errors.


Morxemplum:
Fixed merge issues. Was really trivial stuff.

PR #173, opened 2020-06-20
Expand Cap Color to be more customizable

Morxemplum:
This expands upon the Cap Color code that was originally introduced to now accept color data, such as how it is done when configuring the main color or the colors for your background panels. 

1.92 style colorization is known as "legacy" and can be switched on and off. Also includes one quick fix of C++ code and also updates all of the level styles to have cap color configurations that seem to be most ideal for them.


Morxemplum:
Decided to add the hotfix on this pull request (was gonna do it on a separate pull request but Github just refused to cooperate)

PR #174, opened 2020-06-21
Fix timer desync

kiwec:
Incrementing currentTime every frame leads to loss of precision, which is sometimes significant (the timer slows down to a crawl after 1000 seconds, depending on your framerate).

Storing the start timestamp and comparing it to the current timestamp ensures we don't drift off the correct time.


include/SSVOpenHexagon/Core/HGStatus.hpp
```diff
@@ -4,12 +4,19 @@
 
 #pragma once
 
+#include <sys/time.h>
```
SuperV1234, line 7: I'm going to be a pain and ask you if you can change this to use `<chrono>`, which is the portable way of performing time-related operations in C++11 and above. `<sys/time.h>` is not portable.

https://en.cppreference.com/w/cpp/chrono/

include/SSVOpenHexagon/Core/HGStatus.hpp
```diff
@@ -29,6 +36,29 @@ struct HexagonGameStatus
...
     ssvu::ObfuscatedValue<float> lostFrames{0};
+
+    HexagonGameStatus() {
+        struct timeval time;
+        gettimeofday(&time, NULL);
+        globalTms = time.tv_sec * 1000000 + time.tv_usec;
+        incrementTms = globalTms;
+    }
+
+    void resetIncrementTime() {
```
SuperV1234, line 47: Is this unused?

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -181,8 +183,7 @@ void HexagonGame::updateTimeStop(FT mFT)
 {
     if(status.timeStop <= 0)
     {
-        status.currentTime += ssvu::getFTToSeconds(mFT);
-        status.incrementTime += ssvu::getFTToSeconds(mFT);
+        status.updateTime();
```
Morxemplum, line 186: Check to make sure this new implementation doesn't break u_haltTime, which will use status.timeStop

kiwec:
I have updated the code to use `<chrono>`, but I'm struggling with duration conversions. The time pausing doesn't work because of this. Anyone wants to finish this PR ?

kiwec:
Fixed the issues related to the `<chrono>` conversion, thanks @Morxemplum for the help.

Also fixed the timer starting before rotating, and added time pausing support.

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -2,6 +2,8 @@
 // License: Academic Free License ("AFL") v. 3.0
 // AFL License page: http://opensource.org/licenses/AFL-3.0
 
+#include <sys/time.h>
+
```
Morxemplum, line 6: Remove this please

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -45,7 +45,7 @@ void HexagonGame::initLua_Utils()
     lua.writeVariable("u_isKeyPressed",
         [this](int mKey) { return window.getInputState()[KKey(mKey)]; });
     lua.writeVariable(
-        "u_haltTime", [this](float mDuration) { status.timeStop = mDuration; });
+        "u_haltTime", [this](float mDuration) { status.pauseTime(mDuration); });
```
Morxemplum, line 48: I think you are pausing this in seconds notation. This is supposed to be in frames. Just quickly divide this by 60 and you're good.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -126,12 +126,12 @@ void HexagonGame::initLua_MainTimeline()
 void HexagonGame::initLua_EventTimeline()
 {
     lua.writeVariable("e_eventStopTime", [this](float mDuration) {
-        eventTimeline.append<Do>([=, this] { status.timeStop = mDuration; });
+        eventTimeline.append<Do>([=, this] { status.pauseTime(mDuration); });
```
Morxemplum, line 129: Same comment applies here.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -126,12 +127,13 @@ void HexagonGame::initLua_MainTimeline()
...
     lua.writeVariable("e_eventStopTime", [this](float mDuration) {
-        eventTimeline.append<Do>([=, this] { status.timeStop = mDuration; });
+        eventTimeline.append<Do>(
+            [=, this] { status.pauseTime(ssvu::getSecondsToFT(mDuration)); });
     });
 
     lua.writeVariable("e_eventStopTimeS", [this](float mDuration) {
         eventTimeline.append<Do>(
-            [=, this] { status.timeStop = ssvu::getSecondsToFT(mDuration); });
+            [=, this] { status.pauseTime(ssvu::getSecondsToFT(mDuration)); });
```
Morxemplum, line 136: I mean since your pause time code is already written in seconds, you can get rid of this conversion entirely.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -44,8 +44,9 @@ void HexagonGame::initLua_Utils()
     });
     lua.writeVariable("u_isKeyPressed",
         [this](int mKey) { return window.getInputState()[KKey(mKey)]; });
-    lua.writeVariable(
-        "u_haltTime", [this](float mDuration) { status.timeStop = mDuration; });
+    lua.writeVariable("u_haltTime", [this](float mDuration) {
+        status.pauseTime(ssvu::getSecondsToFT(mDuration));
+    });
```
Morxemplum, line 49: You are doing the exact opposite of what I've told you to do. You are MULTIPLYING it by the framerate, not DIVIDING it. Use getFTToSeconds to convert frame time to seconds.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -126,12 +127,13 @@ void HexagonGame::initLua_MainTimeline()
 void HexagonGame::initLua_EventTimeline()
 {
     lua.writeVariable("e_eventStopTime", [this](float mDuration) {
-        eventTimeline.append<Do>([=, this] { status.timeStop = mDuration; });
+        eventTimeline.append<Do>(
+            [=, this] { status.pauseTime(ssvu::getSecondsToFT(mDuration)); });
     });
```
Morxemplum, line 132: Same comment I've made on u_haltTime applies here

Morxemplum:
Just tried it with these fixes. It solves the freezing timer issue but I'm getting Jumps in the time now. This is an avenger levels threat.

Morxemplum:
We finally got this one down. This one was a toughie.

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -221,6 +221,8 @@ void HexagonGame::sideChange(unsigned int mSideNumber)
 
 void HexagonGame::checkAndSaveScore()
 {
+    float time = status.getTimeSeconds();
```
SuperV1234, line 224:
```suggestion
    const float time = status.getTimeSeconds();
```

include/SSVOpenHexagon/Core/HGStatus.hpp
```diff
@@ -4,15 +4,29 @@
...
 
 namespace hg
 {
 struct HexagonGameStatus
 {
-    ssvu::ObfuscatedValue<float> currentTime{0.f};
-    float incrementTime{0};
-    float timeStop{100};
+private:
+    // Time point of the current game loop
```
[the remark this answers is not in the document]
Morxemplum: I think that means the current time of the level.

kiwec:
This is dumb

src/SSVOpenHexagon/Core/HGStatus.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+using namespace hg;
+
+float HexagonGameStatus::getIncrementTimeSeconds()
+{
+    if(isTimePaused())
+    {
+        auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+            pauseTp - incrementTp);
+        return static_cast<float>(ms.count()) / 1000.f;
```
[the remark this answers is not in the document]
Morxemplum: Maybe this could be casted to an integer

Morxemplum:
Reopening this pull request. I think a lot of Vee's changes can easily be done. It may be annoying but I think he's doing this because he wants the timer to be as accurate as possible.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -126,12 +127,12 @@ void HexagonGame::initLua_MainTimeline()
 void HexagonGame::initLua_EventTimeline()
 {
     lua.writeVariable("e_eventStopTime", [this](float mDuration) {
-        eventTimeline.append<Do>([=, this] { status.timeStop = mDuration; });
+        eventTimeline.append<Do>(
+            [=, this] { status.pauseTime(ssvu::getFTToSeconds(mDuration)); });
```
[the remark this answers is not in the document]
Morxemplum: No. This Lua function specifically takes in a parameter of frame time notation, and the pauseTime method that he coded accepts seconds, so a conversion has to be made in order to get it to the correct notation. The code he wrote here is correct. 

If you want to actively use seconds, use e_eventStopTimeS.

include/SSVOpenHexagon/Core/HGStatus.hpp
```diff
@@ -4,15 +4,29 @@
...
+    std::chrono::time_point<std::chrono::steady_clock> lastTp;
+
+    // When we started playing the level
+    std::chrono::time_point<std::chrono::steady_clock> startTp;
+
+    // When the last increment happened
+    std::chrono::time_point<std::chrono::steady_clock> incrementTp;
```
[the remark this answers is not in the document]
Morxemplum: Pretty much. But in the case that a force increment or one of the event pauses is called, those are going to be used. It's best to have them defined then to not have them there when they are going to be used.

include/SSVOpenHexagon/Core/HGStatus.hpp
```diff
@@ -4,15 +4,29 @@
...
+
+    // When the timer was last paused, and for how long
+    std::chrono::time_point<std::chrono::steady_clock> pauseTp;
+    std::chrono::milliseconds pauseDuration{100ms};
```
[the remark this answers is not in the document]
Morxemplum: This might be a small delay for the start of the level. I'm not sure.

Morxemplum:
Left comments on some of Vee's questions.

include/SSVOpenHexagon/Core/HGStatus.hpp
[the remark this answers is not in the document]
SuperV1234:
```suggestion
    // Current time of the level
```

Morxemplum:
Tested again. Achieved a 11 minute record and tested it against a timecode with no error (except for errors irrelevant to the code itself). It really is much more accurate than the previous timer. https://www.youtube.com/watch?v=4ZGs1_Fx0IA

PR #178, opened 2020-06-27
Scrap l_maxInc and use different level checks to stop impossible levels

### LINKED ISSUE #175: Scrap l_maxInc and replace it with different level checks
Between RC2 and RELEASE Build 6.16.2020, there has been an addition to level coding, l_maxInc, that caps the amount of times that a level can increment per attempt. The point of this function was supposed to cap the incrementation of levels to ensure that they do not reach a point that exceeds human limits. 

However, there have been major oversights that have not been taken into account when implementing this. Many levels in Open Hexagon rely on increments to ensure that the level doesn't become stale and monotone, and using l_maxInc stops this from happening, which means that levels are less based on the difficulty and more on endurance / patience. I have seen this myself when I reached 140 seconds on Pointless and the level stops incrementing. The sides never change and it felt like a patience challenge. 

This is even more detrimental on hypercube levels like Incongruence, which relies on increments for changing ranges. 

I feel like there are better ways to ensure that a level can't go beyond human limits (such as capping the max speed), so that's why I'm raising this as an issue, and it can be resolved in the upcoming release of Open Hexagon 2.

Morxemplum:
Fixes #175 

Levels have been turned to now have speedMaxes and DelayMins (with DelayMax added as well, but wasn't needed). l_maxInc is no longer an existing Lua function and shouldIncrement has been removed completely.


src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -182,11 +182,8 @@ void HexagonGame::incrementDifficulty()
 {
     assets.playSound("levelUp.ogg");
 
-    if(levelStatus.shouldIncrement())
-    {
-        levelStatus.rotationSpeed += levelStatus.rotationSpeedInc *
-                                     ssvu::getSign(levelStatus.rotationSpeed);
-    }
+    levelStatus.rotationSpeed +=
+        levelStatus.rotationSpeedInc * getSign(levelStatus.rotationSpeed);
```
SuperV1234, line 186:
```suggestion
        levelStatus.rotationSpeedInc * ssvu::getSign(levelStatus.rotationSpeed);
```

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -204,12 +201,24 @@ void HexagonGame::sideChange(unsigned int mSideNumber)
...
+        levelStatus.speedMult = speedMax;
     }
 
+    // Clamp delay between delayMin and delayMax
+    const auto& delayMin(levelStatus.delayMin);
+    const auto& delayMax(levelStatus.delayMax);
+    if (levelStatus.delayMult > delayMax && delayMax > 0.f) 
+        levelStatus.delayMult = delayMax;
+    else if (levelStatus.delayMult < delayMin && delayMin > 0.f)
+        levelStatus.delayMult = delayMin;
```
SuperV1234, line 220:
```suggestion
    const auto& delayMin(levelStatus.delayMin);
    const auto& delayMax(levelStatus.delayMax);
    if (levelStatus.delayMult > delayMax && delayMax > 0.f) 
    {
        levelStatus.delayMult = delayMax;
    }
    else if (levelStatus.delayMult < delayMin && delayMin > 0.f)
    {
        levelStatus.delayMult = delayMin;
    }
```

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -204,12 +201,24 @@ void HexagonGame::sideChange(unsigned int mSideNumber)
...
+    // Cap speed at speed max (with difficulty in mind). If speed max is zero, disregard the cap.
+    const auto& speedMax(levelStatus.speedMax);
+    if (levelStatus.speedMult * difficultyMult > speedMax && speedMax > 0.f)
     {
-        levelStatus.speedMult += levelStatus.speedInc;
-        levelStatus.delayMult += levelStatus.delayInc;
+        levelStatus.speedMult = speedMax;
     }
 
+    // Clamp delay between delayMin and delayMax
```
SuperV1234, line 214:
```suggestion
    // Clamp delay between `delayMin` and `delayMax` if they're non-zero.
```

Morxemplum:
Applied. :)

PR #180, opened 2020-06-28
Improve Common Library and make small adjustments to Cube levels

Morxemplum:
This is the last PR that I plan on doing before I start making another RELEASE build for the general community to try out. This PR focuses a lot more on the front end of Open Hexagon and a lot less on the source code. While both Cube and Hypercube have changes, this PR is a lot more focused on the Cube pack.

Changes:
- Added pSpiral as a common pattern, which is a spiral of cWall.
      - This pattern is seen in Baby Steps, Pointless, Flattering Shape, and Golden Ratio
- pMirrorSpiralDouble has now been fixed so now it functions properly on polygons other than the Hexagon
      - As a result, it has now been added into Euclidean Plane Crash
- Pattern key sequences now reshuffle every time the key sequence loops. This ensures that pattern order doesn't become predictable. (This is done on both Cube and Hypercube levels where key sequences are used)
- Made optimizations to a few functions found in common: getRandomDir() and getSideDistance(). They now have more efficient implementations.
- Fixed the timing of some of the events on Baby Steps
- Stripped the pulse from Golden Ratio completely (to reflect a bit more like the 1.92 counterpart, because imo the 1.92 version was better). Also lowered the initial rotation a bit more to make it less noticeable. (It's not 0 so a rotation increment can happen. Possible bug?)


_RELEASE/Packs/cube/Scripts/Levels/babysteps.lua
```diff
@@ -42,7 +43,7 @@ function onLoad()
 	m_messageAddImportant("avoid the walls!", 130)
 	e_eventStopTimeS(6) e_eventWaitS(6)
 
-	e_eventStopTimeS(3) e_eventWaitUntilS(12)
+	e_eventWaitUntilS(10) e_eventStopTimeS(5)
```
SuperV1234, line 46:
```suggestion
	e_eventWaitUntilS(10) 
	e_eventStopTimeS(5) 
```

_RELEASE/Packs/cube/Scripts/Levels/goldenratio.lua
```diff
@@ -35,10 +36,10 @@ function onInit()
...
 
 	l_setPulseMin(60)
-	l_setPulseMax(87)
-	l_setPulseSpeed(1.2)
-	l_setPulseSpeedR(1)
-	l_setPulseDelayMax(12.9)
+	--l_setPulseMax(87)
+	--l_setPulseSpeed(1.2)
+	--l_setPulseSpeedR(1)
+	--l_setPulseDelayMax(12.9)
```
SuperV1234, line 42:
```suggestion
	-- l_setPulseMax(87)
	-- l_setPulseSpeed(1.2)
	-- l_setPulseSpeedR(1)
	-- l_setPulseDelayMax(12.9)
```

PR #183, opened 2020-06-28
Fix Bugs with speedMax and delay bounds in different difficulties

Morxemplum:
I had an actual brain fart for this one. I am so shocked I missed such obvious mistakes.

This fixes the speedMax, delayMin, and delayMax problems on different difficulties.

### DIFF
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -262,12 +262,14 @@ public:
 
     [[nodiscard]] float getSpeedMultDM() const noexcept
     {
-        return levelStatus.speedMult * (std::pow(difficultyMult, 0.65f));
+        const auto res = levelStatus.speedMult * (std::pow(difficultyMult, 0.65f));
+        return res < levelStatus.speedMax ? res : levelStatus.speedMax;
     }
 
     [[nodiscard]] float getDelayMultDM() const noexcept
     {
-        return levelStatus.delayMult / (std::pow(difficultyMult, 0.10f));
+        const auto res = levelStatus.speedMult / (std::pow(difficultyMult, 0.10f))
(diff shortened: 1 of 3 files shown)


include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -262,12 +262,14 @@ class HexagonGame
 
     [[nodiscard]] float getSpeedMultDM() const noexcept
     {
-        return levelStatus.speedMult * (std::pow(difficultyMult, 0.65f));
+        const auto res = levelStatus.speedMult * (std::pow(difficultyMult, 0.65f));
+        return (res < levelStatus.speedMax && levelStatus.speedMax > 0.f) ? res : levelStatus.speedMax;
```
SuperV1234, line 266: If `levelStatus.speedMax` is zero, this line seems to always return `levelStatus.speedMax`, which seems wrong. Is that intended? Or should it be something like:

```cpp
if (levelStatus.speedMax <= 0.f)
{
   return res; 
}

return std::clamp(res, 0.f, levelStatus.speedMax);
```

?

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -262,12 +262,14 @@ class HexagonGame
...
     }
 
     [[nodiscard]] float getDelayMultDM() const noexcept
     {
-        return levelStatus.delayMult / (std::pow(difficultyMult, 0.10f));
+        const auto res = levelStatus.delayMult / (std::pow(difficultyMult, 0.10f));
+        return std::clamp(res, levelStatus.delayMin, levelStatus.delayMax);
```
SuperV1234, line 272: Should be consistent with the above function, and handle special zero cases for `delayMin` and `delayMax` here?

PR #185, opened 2020-06-29
Additional functions for lua and other fixes/additionals

Zly-u:
-Collisions now are processed separately from the graphics for correct calculations since I had to move skew into graphics.
-Fixed Wall's being inaccurate at stopping on the central shape.
-Changed Wall's logic to be able to make future functionality such as smooth side change and walls adaptation to side change.
-Changed graphics draw logic in general. (not entirely though)
-Capped settable `side` param for walls from 0 to the number of level's sides only, so there will be no calculation errors if someone decides to set a huge value in there.
-applied a `delta time` to rotations decrease when the player dies with the help of `std::pow`.
-refactored the code a bit.
-changed the update order for the Level so style data or something else doesn't override Lua commands.
-fixed background in level preview in the menu to black but level's last color in the style's color table.

`Added`:
-s_BGRotationOffset(val)
-s_BGTileRadius(val)
-s_ColorPosOffset(val)
-s_setSkew(x, y), s_getSkew(): x, y
-l_setFieldPos(x, y), l_getFieldPos(): x, y
-u_swapPlayer()


Morxemplum:
I did my very best on trying to solve the merge conflicts. These ones were very weird. If there's anything wrong I'm sorry.

Morxemplum:
I dunno about this one chief. This one seems kinda buggy.

Morxemplum:
The new lua functions need documentation for sure.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -375,6 +380,14 @@ void HexagonGame::initLua_LevelControl()
...
             "`$1`. Tracked variables are displayed in game, below the game "
             "timer.");
+  
+    lua.writeVariable("l_setFieldPos", [this](float mX, float mY) {
+        levelStatus.fieldPos = {mX, mY};
+    });
+
+    lua.writeVariable("l_getFieldPos", [this] {
+        return std::make_tuple(levelStatus.fieldPos.x, levelStatus.fieldPos.y);
+    });
```
Morxemplum, line 390: Use lsVar for this?

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -489,6 +506,14 @@ void HexagonGame::initLua_StyleControl()
...
             "Set the color of the center polygon to match the  style "
             "color with index `$0`.");
+  
+    lua.writeVariable("s_setSkew", [this](float mX, float mY) {
+        styleData.skew = {mX, mY};
+    });
+
+    lua.writeVariable("s_getSkew", [this] {
+        return std::make_tuple(styleData.skew.x, styleData.skew.y);
+    });
```
Morxemplum, line 516: Use sdVar?

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -151,6 +151,8 @@ void HexagonGame::initLua_Utils()
         .doc(
             "Return the current delay multiplier, adjusted for the chosen "
             "difficulty multiplier.");
+  
+    lua.writeVariable("u_swapPlayer", [this]() { playerSwap(false); });
```
Morxemplum, line 155: Use addLuaFn, so that way you can add documentation to these functions.

include/SSVOpenHexagon/Data/StyleData.hpp
```diff
@@ -7,6 +7,7 @@
 #include "SSVOpenHexagon/Global/Common.hpp"
 #include "SSVOpenHexagon/Data/ColorData.hpp"
 #include "SSVOpenHexagon/Data/CapColor.hpp"
+#include "SSVOpenHexagon/Core/HGStatus.hpp"
```
SuperV1234, line 10: We can avoid a physical dependency on this header file by forward-declaring `HexagonGameStatus`. This usually speeds up compilation.

src/SSVOpenHexagon/Data/StyleData.cpp
```diff
@@ -7,8 +7,17 @@
...
+#include "SSVOpenHexagon/Core/HGStatus.hpp"
+#include "SSVOpenHexagon/Utils/Utils.hpp"
 #include "SSVOpenHexagon/Utils/FastVertexVector.hpp"
 
+using namespace std;
+using namespace sf;
+using namespace ssvs;
+using namespace hg::Utils;
+using namespace ssvu;
+using namespace ssvuj;
```
SuperV1234, line 19: This is considered bad practice - namespaces help to understand where the symbols we're using are coming from, especially in a project with many libraries like Open Hexagon.

src/SSVOpenHexagon/Data/StyleData.cpp
```diff
@@ -123,18 +134,20 @@ void StyleData::computeColors()
...
+        ssvu::rotate(currentColors, begin(currentColors) +
+                                        currentSwapTime / (maxSwapTime / 2.f) +
+                                        colorPosOffset % levelStatus.sides);
     }
 }
 
-void StyleData::drawBackground(sf::RenderTarget& mRenderTarget,
-    const sf::Vector2f& mCenterPos, const LevelStatus& levelStatus) const
+void StyleData::drawBackground(RenderTarget& mRenderTarget,
+    const sf::Vector2f& mCenterPos, LevelStatus& levelStatus,
```
SuperV1234, line 144: Not sure why you got rid of the `const` on `levelStatus` here. Immutability is good to avoid mistakes and help readability

include/SSVOpenHexagon/Core/HGStatus.hpp
```diff
@@ -47,6 +47,7 @@ struct HexagonGameStatus
     float flashEffect{0};
     float radius{75};
     float fastSpin{0};
+    float camFastSpin{0};
```
SuperV1234, line 50: Unused?

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -250,11 +258,21 @@ class HexagonGame
         return game;
     }
 
+    sf::Vector2f getFieldPos() const noexcept
```
SuperV1234, line 261: It's good practice to mark functions whose return value should never be ignored as `[[nodiscard]]`, as accidentally ignoring the return value results in a compiler warning (rather than a silent run-time mistake).

include/SSVOpenHexagon/Data/StyleData.hpp
```diff
@@ -33,8 +34,12 @@ class StyleData
     float pulseIncrement;
     bool huePingPong;
     float maxSwapTime;
+    int colorPosOffset;
```
SuperV1234, line 37: This needs to be initialized, otherwise reading from it would be undefined behavior.

include/SSVOpenHexagon/Data/StyleData.hpp
```diff
@@ -33,8 +34,12 @@ class StyleData
...
+    float bgTileRadius;
+    float bgRotOff;
```
SuperV1234, line 39: Ditto.

include/SSVOpenHexagon/Data/StyleData.hpp
```diff
@@ -33,8 +34,12 @@ class StyleData
...
     float _3dDepth;
     float _3dSkew;
+    sf::Vector2f skew;
```
SuperV1234, line 42: This one should be fine because the default constructor of `sf::Vector2f` zeroes the values out (hopefully).

src/SSVOpenHexagon/Data/StyleData.cpp
Zly-u: I know, not my thing but Synth's merging purpose idk, told him that you are not using them but kept just in case.

include/SSVOpenHexagon/Utils/Utils.hpp
```diff
@@ -102,4 +102,20 @@ void runLuaFunctionIfExists(
     runLuaFunction<T>(mLua, mName, mArgs...);
 }
 
+template <typename T1, typename T2, typename T3>
+inline auto getSkewedVecFromRad(const T1& mRad, const T2& mMag, const T3& mSkew)
```
SuperV1234, line 106: Templates are implicitly `inline`, so the keyword here is redundant.

src/SSVOpenHexagon/Data/StyleData.cpp
Zly-u: huh, well, I actually didn't mean to not set it as constant

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -121,23 +170,45 @@ void CPlayer::drawDeathEffect(HexagonGame& mHexagonGame)
...
 {
+    startPos = mHexagonGame.getFieldPos();
+    sf::Vector2f lastPos{pos};
+    float currentSpeed{speed};
+
+    const float lastAngle{angle};
+    const float radius{mHexagonGame.getRadius()};
+    const int movement{mHexagonGame.getInputMovement()};
+
+    angle += ssvu::toRad(currentSpeed * movement * mFT);
```
SuperV1234, line 210: Why was this moved up? Now `currentSpeed` will lag one frame behind, for example when using focus.

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -337,8 +334,15 @@ void HexagonGame::updateRotation(FT mFT)
             getSign(nextRotation);
         status.fastSpin -= mFT;
     }
-
-    backgroundCamera.turn(nextRotation);
+    levelStatus.rotation += nextRotation;
+    if(levelStatus.rotation >= 360.f)
```
SuperV1234, line 338: Probably won't happen, but what if `nextRotation` is greater than 360?

Zly-u: Will cap it then with modulo, maybe, will try to figure it out actually

Morxemplum:
Please modify your branch so it works with the new extlibs. I cannot compile this at the moment.

Morxemplum:
Alright you just need to fix a few issues. You need to ensure that the death markers are accurate, because when I die my arrow is deep inside a wall. Also, some of the walls are not being drawn onto the background with certain patterns, which is a problem. Also for the new extlibs it for some reason enlargens the bottom hotbar. Fix all of these and I think you may get merged. Don't give up! :)

Morxemplum:
Gonna close this PR. It's already way behind the latest features and this can be broken up into smaller PRs. :)

PR #187, opened 2020-07-02
Reposition Level Text For New Font

### LINKED ISSUE #179: Enlargen In Game Timer to Adjust to New Font
Due to the death of the use of the iconic Imagine Font (Rest In Peace) for licensing issues, we now have to use the new font named Forced Square, which is similar to Imagine but doesn't have the same licensing issues. 

There's only one big issue I have with this font, and it's how this font handles digits. The digits are a lot more "narrow" in width compared to Imagine, and it looks aesthetically unpleasing. I know this is an issue with all places that use numbers, but I think the most critical one is the timer itself. This issue is going to be focusing on slightly enlarging the timer to make it work better with the new font.

I plan on stripping the "Time:" portion completely and show the raw number (because I think it'll be trivial enough to know that it's the timer). And just enlargen the number to make it stand out. This will be challenging because I will have to mess with the GUI code a bit to get this to work, but I think it can be done.

Morxemplum:
Fixes #179 

In addition, it also moves the FPS to the bottom left; the in-game timer has been redone to show the hundreths place no matter whether you are in 10, 100, or 1000 seconds; and the outline code has been redone to take advantage of SFML's native implementation of text outline, saving several draw calls with the text and fixing a lot of bugs. 

I also did test this with different resolutions to ensure no scaling problems. The pictures attached show the difference between 1080p and 480p, with the only noticable differences being the thickness of the outline and the X positioning (which was already a thing before. I just wanted to make sure the Y positioning doesn't mess up). Even changing the resolution in game doesn't cause glitches.

SSVOpenHexagon_2020-07-02_11-28-59
SSVOpenHexagon_2020-07-02_12-39-15


Morxemplum:
Don't mind the X thing in one of the screenshots. That's one of my widgets and has nothing to do with the game graphics.

PR #193, opened 2020-07-10
Strengthen Lua Errors

### LINKED ISSUE #188: Be More Strict On Lua Errors
I'm so annoyed when pack developers gloss over level errors because "it doesn't matter, the level just works." or there's always the trailing errors that print every frame so it either covers up other errors or is just annoying to read in general.

What I plan on doing with this is making Lua Diagnostics more strict by forcing the player back into the menu whenever an error occurs, and an "error" sound will play instead of the "Open Hexagon" sound, which should tell pack developers that something went wrong and they should check the console for details. They have to fix those errors before the level is fully playable.

The only way that you'll be able to bypass this strict move is to enable Debug Mode, which should help give Debug Mode more meaning. 

This should also provide contribution to issue #181, considering the new sound could also make good use there.

Morxemplum:
Fixes #188

Added an error sound to the game (which could be changed at any time if it's not your favorite). All expectations are met within the issue: You get kicked to the menu whenever a Lua error happens, an error sound plays, which means you cannot play the level until the errors are fixed. You can only bypass this by turning on Debug mode (which I now added in the game settings because the only way to do this previously was to run the config override and disable it in the JSON).

Also slightly reformatted the Lua errors to present themselves better and to also show the level that the errors come from (just to lower confusion). 

This feature can definitely be expanded upon in separate PRs (like Gui notices for Lua errors) but I think this is a good start.


_RELEASE/Assets/assets.json
```diff
@@ -21,6 +21,7 @@
 		"go.ogg",
 		"levelUp.ogg",
 		"openHexagon.ogg",
+		"error.ogg",
```
SuperV1234, line 24: What's the license of this sound file? Is it free to use in a commercial project?

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
     template <typename T, typename... TArgs>
     T runLuaFunction(const std::string& mName, const TArgs&... mArgs)
     {
-        return Utils::runLuaFunction<T, TArgs...>(lua, mName, mArgs...);
+        try {
+            return Utils::runLuaFunction<T, TArgs...>(lua, mName, mArgs...);
+        } 
+        catch(std::runtime_error& mError)
```
SuperV1234, line 166:
```suggestion
        catch(const std::runtime_error& mError)
```

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
...
+        {
+            std::cout << "[Lua] Runtime error on \""
+                      << mName
+                      << "\" with level \""
+                      << levelData -> name
```
SuperV1234, line 171:
```suggestion
                      << levelData->name
```

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
...
+                      << "\": \n"
+                      << ssvu::toStr(mError.what()) << "\n"
```
SuperV1234, line 173: Do we need the `toStr` here?

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
     template <typename T, typename... TArgs>
     T runLuaFunction(const std::string& mName, const TArgs&... mArgs)
     {
-        return Utils::runLuaFunction<T, TArgs...>(lua, mName, mArgs...);
+        try {
+            return Utils::runLuaFunction<T, TArgs...>(lua, mName, mArgs...);
+        } 
+        catch(std::runtime_error& mError)
+        {
+            std::cout << "[Lua] Runtime error on \""
```
SuperV1234, line 168:
```suggestion
            std::cout << "[runLuaFunction] Runtime error on \""
```

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
...
     template <typename T, typename... TArgs>
     void runLuaFunctionIfExists(const std::string& mName, const TArgs&... mArgs)
     {
-        Utils::runLuaFunctionIfExists<T, TArgs...>(lua, mName, mArgs...);
+        try {
+            Utils::runLuaFunctionIfExists<T, TArgs...>(lua, mName, mArgs...);
+        }
+        catch(std::runtime_error& mError)
+        {
+            std::cout << "[Lua] Runtime error on \""
```
SuperV1234, line 191:
```suggestion
            std::cout << "[runLuaFunctionIfExists] Runtime error on \""
```

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
...
+                      << mName
+                      << "\" with level \""
+                      << levelData -> name
+                      << "\": \n"
+                      << ssvu::toStr(mError.what()) << "\n"
+                      << std::endl;
+            if (!Config::getDebug())
+            {
+                goToMenu(false, true);
```
SuperV1234, line 177:
```suggestion
                goToMenu(false /* mSendScores */, true /* mError */);
```

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
...
+            std::cout << "[Lua] Runtime error on \""
+                      << mName
+                      << "\" (optional function) with level \""
+                      << levelData -> name
+                      << "\": \n"
+                      << ssvu::toStr(mError.what()) << "\n"
+                      << std::endl;
+            if (!Config::getDebug())
+            {
+                goToMenu(false, true);
```
SuperV1234, line 200:
```suggestion
                goToMenu(false /* mSendScores */, true /* mError */);
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -684,8 +692,20 @@ void MenuGame::setIndex(int mIdx)
     Lua::LuaContext lua;
     initLua(lua);
     Utils::runLuaFile(lua, levelData->luaScriptPath);
-    Utils::runLuaFunction<void>(lua, "onInit");
-    Utils::runLuaFunction<void>(lua, "onLoad");
+    try {
+        Utils::runLuaFunction<void>(lua, "onInit");
+        Utils::runLuaFunction<void>(lua, "onLoad");
+    } catch (std::runtime_error& mError) {
+        std::cout << "[Lua] Runtime error on menu (onInit/onLoad) with level \""
```
SuperV1234, line 699:
```suggestion
        std::cout << "[MenuGame::init] Runtime Lua error on menu (onInit/onLoad) with level \""
```

src/SSVOpenHexagon/Core/main.cpp
```diff
@@ -104,7 +104,7 @@ int main(int argc, char* argv[])
 
     assets->refreshVolumes();
     window.setGameState(mg->getGame());
-    mg->init();
+    mg->init(false);
```
SuperV1234, line 107:
```suggestion
    mg->init(false /* mError */);
```

_RELEASE/Assets/assets.json
Morxemplum: This was a quick sound that I have whipped up in Audacity, where I took a source sound and edited it to make an error sound (it was just a singular ding which I modified). The source sound I got the audio from claims the audio is under Public Domain, which means even if it isn't transformative enough we cannot be sued for copyright infringement because there isn't one.

include/SSVOpenHexagon/Core/HexagonGame.hpp
Morxemplum: I'm not sure. This was in the legacy implementation of reporting Lua errors, and it looks like what() returns a pointer to an array of characters, so if you want to consider this a string, feel free to make the changes.

include/SSVOpenHexagon/Core/HexagonGame.hpp
Morxemplum: Could you amend this a bit so that if you are going to report it for "IfExists" specifically, you could get rid of "(optional function)". That way it's not so redundant.

_RELEASE/Assets/assets.json
SuperV1234: Good enough for me. Cheers.

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -160,13 +160,46 @@ class HexagonGame
...
     {
-        Utils::runLuaFunctionIfExists<T, TArgs...>(lua, mName, mArgs...);
+        try {
+            Utils::runLuaFunctionIfExists<T, TArgs...>(lua, mName, mArgs...);
+        }
+        catch(std::runtime_error& mError)
+        {
+            std::cout << "[Lua] Runtime error on \""
+                      << mName
+                      << "\" (optional function) with level \""
```
SuperV1234, line 193:
```suggestion
                      << "\" with level \""
```

PR #195, opened 2020-07-10
Fix zero rotation issue

### LINKED ISSUE #191: [Bug] Rotation Speed doesn't increment when equal to 0
When rotation speed is 0, no matter what the increment value is, the rotation speed doesn't increment on level up.

I know what is causing this bug (this has to do with the sign function), but I'm raising it as an issue so that way I can remind myself to fix it.

Morxemplum:
Fixes #191.

As a result, I set Golden Ratios beginning speed equal to 0. Now it is how it was intended to be.

### DIFF
--- a/src/SSVOpenHexagon/Core/HexagonGame.cpp
+++ b/src/SSVOpenHexagon/Core/HexagonGame.cpp
@@ -190,8 +190,10 @@ void HexagonGame::incrementDifficulty()
 {
     assets.playSound("levelUp.ogg");
 
+    const int signMult = ssvu::getSign(levelStatus.rotationSpeed) == 0 ? 1 : ssvu::getSign(levelStatus.rotationSpeed);
+
     levelStatus.rotationSpeed +=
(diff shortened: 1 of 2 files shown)


src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -190,16 +190,18 @@ void HexagonGame::incrementDifficulty()
 {
     assets.playSound("levelUp.ogg");
 
+    const int signMult = ssvu::getSign(levelStatus.rotationSpeed) == 0 ? 1 : ssvu::getSign(levelStatus.rotationSpeed);
```
SuperV1234, line 193:
```suggestion
    const float signMult = (levelStatus.rotationSpeed > 0.f) ? 1.f : -1.f;
```

Wouldn't this work and be simpler?

Morxemplum: Fair enough. And it lowers reliance on SSVU.

PR #197, opened 2020-07-10
Fix spirals not randomizing direction in common patterns

Morxemplum:
This was a bug that has been picked up by one of the community members last night, where he noticed that mirror spirals were not randomizing in their direction, but singular spirals did. It turns out that this was happening because "loopDir", the variable that is used to assign the direction of where the spiral goes, was constantly being overwritten by cWallEx, a function called in the spiral functions to create the walls. This pull request fixes this issue by renaming the loopDir variable in cWallEx to something else, so it doesn't overwrite the value of the loopDir in spiral patterns.

I also made it so pSpiral uses cWallEx now instead of just cWall and I also made a ton of variables in common patterns local, to ensure scope safety of a lot of the variables.

### DIFF
--- a/_RELEASE/Packs/cube/Scripts/alternativepatterns.lua
+++ b/_RELEASE/Packs/cube/Scripts/alternativepatterns.lua
@@ -1,11 +1,11 @@
 u_execScript("common.lua")
 
 function pAltMirrorSpiral(mTimes, mExtra)
-	oldThickness = THICKNESS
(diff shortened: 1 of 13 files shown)


Morxemplum:
I went ahead and added scope protection for the rest of the pattern files. And I test every single level and it seems as if every level is unaffected. I think it's safe to say that these changes are harmless.

PR #201, opened 2020-07-12
Add 3D Required as a toggleable requirement for score eligibility on 3D-reliant levels

### LINKED ISSUE #200: Add a 3D requirement for 3D-reliant levels
I'm surprised I haven't made an issue about this earlier, but there are plenty of custom levels in 1.92 packs that like to use the tag [3D Required]. What this tag means is that the main gameplay of the level is heavily reliant on the 3D aspects of the game engine to work as intended. However, while Official Mode is very strict on what scores are legitimate and which ones aren't, one of the things that Official Mode can't really touch that well is whether 3D must be enabled for a score to count. And there are some levels where playing them in 2D or 3D don't make a difference.

For this issue, I plan on adding a Lua function for l_set3DRequired(boolean) and l_get3DRequired() for whether a level must be played with 3D on for a score to be considered legitimate. By default, this will be false. However, if it is true and if a user is not playing on 3D, it will show on the top left their score has been invalidated and say (3D Required). It's a silly addition but I think it is something that pack developers would love to see.

Morxemplum:
Fixes #200 

Does pretty much it asks. Use l_set3dRequired(boolean) and l_get3dRequired. 
I also:
- Made the "Score Invalidated" display parameterized, which means that you can now add many more reasons to invalidate a score. 
- Fixed a major oversight in the game design and I heavily increased the standards for local scores to have the same standards online scores have. I have no idea why it was the case that the standards were so low for local scores, but now you can't just turn off official mode and get away with it.
- Added an experimental level to try out the 3D required yourself.

### DIFF
--- a/src/SSVOpenHexagon/Core/HexagonGame.cpp
+++ b/src/SSVOpenHexagon/Core/HexagonGame.cpp
@@ -256,6 +256,11 @@ void HexagonGame::checkAndSaveScore()
                 << "Not sending score - less than 8 seconds\n";
             return;
         }
+        if (levelStatus.3DRequired && !Config::get3D())
+        {
+            ssvu::lo("hg::HexagonGame::checkAndSaveScore()")
+                << "Not saving score - 3D not enabled on a 3D Required level\n";
+        }
         Online::trySendScore(levelData->id, difficultyMult, time);
     }
 }
--- a/src/SSVOpenHexagon/Core/HexagonGame.cpp
+++ b/src/SSVOpenHexagon/Core/HexagonGame.cpp
@@ -224,11 +224,19 @@ void HexagonGame::checkAndSaveScore()
 {
     const float time = status.getTimeSeconds();
 
-    if(Config::getInvincible())
+    // These are requirements that need to be met for a score to be valid
+    if(status.scoreInvalid || !Config::isEligibleForScore())
     {
         ssvu::lo("hg::HexagonGame::checkAndSaveScore()")
-            << "Not saving score - invincibility on\n";
+            << "Not sending/saving score - not eligible\n"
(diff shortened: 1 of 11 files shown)


src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -243,19 +251,25 @@ void HexagonGame::checkAndSaveScore()
...
+			ssvu::lo("hg::HexagonGame::checkAndSaveScore()")
+                << "Not sending score - connection error\n";
+			return;
+		}
+		if(Online::getServerVersion() > Config::getVersion())
+		{
+			ssvu::lo("hg::HexagonGame::checkAndSaveScore()")
+                << "Not sending score - version mismatch\n";
+			return;
+		}
```
SuperV1234, line 272: I'm not sure why you put these here and removed the check on `if(status.scoreInvalid || !Config::isEligibleForScore())` - can you explain?

Morxemplum: Since I have now moved isEligibleForScore to be called before it checks whether the user is local or not, I had to remove the online checks from it, otherwise the score would otherwise be immediately disqualified; it can't connect to the server because there's no server to connect to.

These checks should only be happening when we know that the player is trying to submit a score online and they aren't a local profile, hence I removed them completely from isEligibleForScore and put them in the online score checking session, which is where they belong. 

And the original if statement is not gone. I decided to split the conditions into two separate if statements because you can't get an uneligibility reason if you manually invalidate the score. Hence a separate error message had to be made.

Morxemplum:
Not really sure what's wrong here. If there's anything wrong with the code feel free to point it out.

PR #214, opened 2020-07-16
Post-Steam Release Bug Fixes

### LINKED ISSUE #210: Rotation is not capping on rotationMax if level has fast spin
Bug is pretty self explanatory. It has to do with a condition in an if statement so it's easy to fix. Will fix this soon.

Morxemplum:
This pull request contains a bundle of bug fixes that I decided to start patching up after the Steam release, thanks to reports in the Discord server.

Bugs that are fixed:
- Rotation doesn't cap on levels with fast spin (Fixes #210)
- mLua freaks out over functions with the style prefix
- The "Pentagon removed" message on Second Dimension and Apeirogon displays on every attempt for high difficulty multipliers
- Second Dimension's Messages are on Euclidean Plane Crash. I don't know if this was intended or not, but I am assuming it wasn't.
- Music plays when No Music is toggled on.
- Fixed message logic for enableSwapIfSpeedGEThan. Not a bug per se but people were getting picky about it.
- Antialiasing doesn't apply automatically on startup. It needs to applied manually through settings.

Bugs that aren't fixed:
- Mouse cursor is visible even though it's disabled. I tried my best to figure how to fix this and even changed up the implementation a bit (and now the option is ignored when the game is windowed), but I don't know if this is a bug that is within our reach. Might be a bug with SFML.


src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -254,7 +257,7 @@ void HexagonGame::incrementDifficulty()
     levelStatus.rotationSpeed *= -1.f;
 
     const auto& rotationSpeedMax(levelStatus.rotationSpeedMax);
-    if(status.fastSpin < 0 && abs(levelStatus.rotationSpeed) > rotationSpeedMax)
+    if(abs(levelStatus.rotationSpeed) > rotationSpeedMax)
```
SuperV1234, line 260: This is the one that I do not understand. This change prevents a "fast spin" from (temporarily) going over the rotation speed max. However, `status.fastSpin` should eventually go back down to `0` due to `status.fastSpin -= mFT;` in `HexagonGame::updateRotation`, so the rotation cap should be applied after the "fast spin" is over.

Maybe what we need is to change `< 0` to `<= 0` to avoid the rare case where fast spin gets decremented exactly to `0`?

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -669,13 +669,22 @@ void MenuGame::initLua(Lua::LuaContext& mLua)
 
     mLua.writeVariable("s_setPulseInc",
         [this](float mValue) { styleData.pulseIncrement = mValue; });
+    
+    mLua.writeVariable("s_setPulseIncrement",
+        [this](float mValue) { styleData.pulseIncrement = mValue; });
```
SuperV1234, line 674: This should already be automatically generated by:

```cpp
sdVar("PulseIncrement", &StyleData::pulseIncrement);
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -669,13 +669,22 @@ void MenuGame::initLua(Lua::LuaContext& mLua)
...
 
     mLua.writeVariable("s_setHueInc",
         [this](float mValue) { styleData.hueIncrement = mValue; });
+    
+    mLua.writeVariable("s_setHueIncrement",
+        [this](float mValue) { styleData.hueIncrement = mValue; });
```
SuperV1234, line 680: Ditto, but by 

This should already be automatically generated by:

```cpp
sdVar("HueIncrement", &StyleData::hueIncrement);
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -669,13 +669,22 @@ void MenuGame::initLua(Lua::LuaContext& mLua)
...
 
     mLua.writeVariable(
         "s_getHueInc", [this] { return styleData.hueIncrement; });
 
+    mLua.writeVariable(
+        "s_getHueIncrement", [this] { return styleData.hueIncrement; });
```
SuperV1234, line 686: Ditto.

src/SSVOpenHexagon/Core/HexagonGame.cpp
Morxemplum: The fast spin only goes over the cap for a specific amount of time before it settles to the rotation speed cap. I've tested this in Golden Ratio for reference. If you don't want to count fast spin to go over the speed cap (as if you want to kill it completely), then feel free to make a suggestion and I'll accept it.

src/SSVOpenHexagon/Core/MenuGame.cpp
Morxemplum: This isn't the case at all for any of the lsVar or sdVar functions. I did this because it's an alias for PulseInc and I wanted to get all the nooks and crannies. sdVar doesn't add any functions to mLua, and this is proven because a Discord member reported this and shows mLua reports errors at style functions, despite them being valid functions.

SuperV1234: I failed to realize this was in `MenuGame.cpp`, sorry.

src/SSVOpenHexagon/Core/HexagonGame.cpp
SuperV1234: OK, my misunderstanding was that I assumed the code above to be executed every frame. But I think that is more reasonable, so how about we keep the condition as:

```cpp
if(status.fastSpin <= 0 && abs(levelStatus.rotationSpeed) > rotationSpeedMax))
```

and move the entire `if` statement in the `updateRotation` function?

PR #216, opened 2020-07-16
New lua style function `s_setColorOffset(`

Zly-u:
Added a style function for offsetting tile colors of the background manually.
So, people who would like to make a "rotating" styles or whatever don't need to make hundreds of style files.


src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -450,6 +450,11 @@ void HexagonGame::initLua_StyleControl()
     sdVar("3dPulseSpeed", &StyleData::_3dPulseSpeed);
     sdVar("3dPerspectiveMult", &StyleData::_3dPerspectiveMult);
 
+    addLuaFn("s_setColorOffset",
+        [this](int mValue) {styleData.BGColorOffset = mValue;})
+        .arg("value")
+        .doc("Offsets the colors of the background tiles.");
```
SuperV1234, line 456:
```suggestion
        .doc("Offsets the colors of the background tiles by `$0`.");
```

PR #224, opened 2020-07-18
Fix Mouse Hide through config.

Zly-u:
Fixed Mouse hide issue. It's based on how SFML work with commands, in which order and how exactly they are called.

So, better sometimes to check for more info about commands that seem to not work.


Zly-u:
> 
> 
> Thank goodness this is finally fixed. Yeah SFML is pretty weird, but this took a while to figure out. At least its fixed now. :)
> 
> I approve it, but Vee always has the final say with this sort of stuff.

Boi I fixed it in 10 min.

Morxemplum:
No need to rub it in my face

Zly-u:
xd.
Also hi Alpha with your notifications. o/

PR #226, opened 2020-07-18
Feature/221 show difficulty multiplier on steam and discord

### LINKED ISSUE #221: [Must Have] - Show difficulty multiplier on Steam and Discord
Currently, when someone is playing a level, it is displayed as so on Steam and Discord:
image

That's cool that they are playing Apeirogon, but how do we know which difficulty multiplier they are playing on? We should show this so friends can make fun of them for playing x0.5, and showing off their impressive skills on >x1.5, or whatever.

Format:
**Apeirogon [x1.0] - 12.345s**

A/C:
1. Show the current difficulty multiplier on Steam and Discord.
2. Current time is rounded to 3 decimal places.

**_Priority: Must Have,_**

AlphaPromethium:
Closes #221.

This enables the ability to show the current difficulty multiplier selected on Steam and Discord as so:
image

Levels that do not have selectable difficulty multipliers will just display as [x1.0].

It doesn't update for Steam at the moment, but I think that is because it's using the `richpresence.txt` released on Steam, however the level name and time formatting is updated. **If this is meant to update locally as well then do let me know how this should be fixed!**

### DIFF
--- a/include/SSVOpenHexagon/Core/Discord.hpp
+++ b/include/SSVOpenHexagon/Core/Discord.hpp
@@ -32,7 +32,7 @@ public:
     bool run_callbacks();
 
     bool set_rich_presence_in_menu();
-    bool set_rich_presence_in_game(std::string str_status);
+    bool set_rich_presence_in_game(const std::string& str_status);
 };
 
 } // namespace hg::Discord
--- a/include/SSVOpenHexagon/Core/Discord.hpp
+++ b/include/SSVOpenHexagon/Core/Discord.hpp
@@ -32,7 +32,7 @@ public:
     bool run_callbacks();
 
     bool set_rich_presence_in_menu();
-    bool set_rich_presence_in_game(std::string_view level_name, float time);
+    bool set_rich_presence_in_game(std::string str_status);
 };
 
 } // namespace hg::Discord
--- a/misc/richpresence.txt
+++ b/misc/richpresence.txt
@@ -4,7 +4,7 @@
     {
 	    "tokens"
 	    {
-	    	"#InGame"	"%levelname%: %time%s"
+	    	"#InGame"	"%levelname% [x%difficultymult%] - %time%s"
 	    	"#InMenu"	"In menu"
 	    }
     }
--- a/src/SSVOpenHexagon/Core/Discord.cpp
+++ b/src/SSVOpenHexagon/Core/Discord.cpp
@@ -101,17 +101,16 @@ bool discord_manager::set_rich_presence_in_menu()
     return true;
 }
 
-bool discord_manager::set_rich_presence_in_game(std::string str_status)
+bool discord_manager::set_rich_presence_in_game(const std::string &str_status)
 {
     if(!_initialized)
     {
         return false;
     }
 
-
     discord::Activity activity{};
     activity.SetState("In game");
-    activity.SetDetails(buf.data());
+    activity.SetDetails(str_status.data());
     _core->ActivityManager().UpdateActivity(activity, [](discord::Result r) {
         if(r != discord::Result::Ok)
         {
--- a/src/SSVOpenHexagon/Core/Discord.cpp
+++ b/src/SSVOpenHexagon/Core/Discord.cpp
@@ -6,6 +6,8 @@
 
 #include <SSVUtils/Core/Log/Log.hpp>
 
+#include <math.h>
+
 #include "discord/discord.h"
 
 namespace hg::Discord
@@ -99,8 +101,7 @@ bool discord_manager::set_rich_presence_in_menu()
     return true;
 }
 
-bool discord_manager::set_rich_presence_in_game(
-    std::string_view level_name, float time)
+bool discord_manager::set_rich_presence_in_game(std::string str_status)
 {
     if(!_initialized)
     {
@@ -109,7 +110,7 @@ bool discord_manager::set_rich_presence_in_game(
 
     static std::string buf;
     buf.clear();
-    buf = std::string(level_name) + " - " + std::to_string(time) + "s";
+    buf = str_status;
 
     discord::Activity activity{};
     activity.SetState("In game");
--- a/src/SSVOpenHexagon/Core/Discord.cpp
+++ b/src/SSVOpenHexagon/Core/Discord.cpp
@@ -108,9 +108,6 @@ bool discord_manager::set_rich_presence_in_game(std::string str_status)
         return false;
     }
 
-    static std::string buf;
-    buf.clear();
-    buf = str_status;
 
     discord::Activity activity{};
     activity.SetState("In game");
--- a/src/SSVOpenHexagon/Core/HGUpdate.cpp
+++ b/src/SSVOpenHexagon/Core/HGUpdate.cpp
@@ -48,8 +48,8 @@ void HexagonGame::update(ssvu::FT mFT)
 
     std::string nameStr = levelData->name;
     nameFormat(nameStr);
-    std::string diffStr = diffFormat(difficultyMult);
-    std::string timeStr = timeFormat(status.getTimeSeconds());
+    const std::string diffStr = diffFormat(difficultyMult);
+    const std::string timeStr = timeFormat(status.getTimeSeconds());
 
     // Presence formatter
     std::string presenceStr = nameStr + " [x" + diffStr + "] - " + timeStr + "s";
--- a/src/SSVOpenHexagon/Core/HGUpdate.cpp
+++ b/src/SSVOpenHexagon/Core/HGUpdate.cpp
@@ -25,14 +25,28 @@ void HexagonGame::update(ssvu::FT mFT)
 {
     mFT *= Config::getTimescale();
 
-    // TODO: refactor to avoid repetition, and truncate floating point number
-    // TODO: also show best record (here) and last run + best record (in menu)
-    steamManager.set_rich_presence_in_game(
-        levelData->name, status.getTimeSeconds());
+    // TODO: show best record (here) and last run + best record (in menu)
+
+    // Name formatter
+    std::string nameStr = std::string(levelData->name);
+    nameStr[0] = std::toupper(nameStr[0]);
+
+    // Difficulty multipler formatter
+    std::string diffStr = std::to_string(HexagonGame::difficultyMult);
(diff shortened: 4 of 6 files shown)


misc/richpresence.txt
```diff
@@ -1,11 +1,11 @@
...
-	    }
-    }
-}
+"lang"
+{
+	"english"
+    {
+	    "tokens"
+	    {
+	    	"#InGame"	"%levelname% [x%difficultymult%] - %time%s"
```
AlphaPromethium, line 7: There is also an issue where the time takes a long time for it to get updated on Steam. This always seems to update when the player dies, which makes me believe it's not updating because it's constantly sending requests to the Steam API every frame.

AlphaPromethium: A card has been raised for this—#227.

src/SSVOpenHexagon/Core/Discord.cpp
```diff
@@ -1,127 +1,128 @@
...
+bool discord_manager::set_rich_presence_in_game(std::string str_status)
+{
+    if(!_initialized)
+    {
+        return false;
+    }
+
+    static std::string buf;
+    buf.clear();
+    buf = str_status;
```
SuperV1234, line 113: The point of the `static` buffer was to provide some storage to do string manipulation with `string_view`, but since you switched to a `string` there's no need for it anymore. 

```suggestion
```

SuperV1234: GitHub doesn't allow me to add a suggestion, but in line `117` you can just say `str_status.data()` instead on `buf.data()`.

include/SSVOpenHexagon/Core/Discord.hpp
```diff
@@ -1,38 +1,38 @@
...
+    discord_manager(const discord_manager&) = delete;
+    discord_manager& operator=(const discord_manager&) = delete;
+
+    discord_manager(discord_manager&&) = delete;
+    discord_manager& operator=(discord_manager&&) = delete;
+
+    bool run_callbacks();
+
+    bool set_rich_presence_in_menu();
+    bool set_rich_presence_in_game(std::string str_status);
```
SuperV1234, line 35: Take strings by `const&` to avoid unnecessary copies.

```suggestion
    bool set_rich_presence_in_game(const std::string& str_status);
```

src/SSVOpenHexagon/Core/Steam.cpp
```diff
@@ -196,16 +196,16 @@ bool steam_manager::set_rich_presence_in_menu()
 }
 
 bool steam_manager::set_rich_presence_in_game(
-    std::string_view level_name, float time)
+    std::string level_name_format, std::string difficulty_mult_format, std::string time_format)
```
SuperV1234, line 199: AFAICS these could all be `std::string_view` instances, which are very lightweight and cheap to copy. Any reason to make them `std::string` instead? 

```suggestion
    std::string_view level_name_format, std::string_view difficulty_mult_format, std::string_view time_format)
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -25,14 +25,28 @@ void HexagonGame::update(ssvu::FT mFT)
...
     mFT *= Config::getTimescale();
 
-    // TODO: refactor to avoid repetition, and truncate floating point number
-    // TODO: also show best record (here) and last run + best record (in menu)
-    steamManager.set_rich_presence_in_game(
-        levelData->name, status.getTimeSeconds());
+    // TODO: show best record (here) and last run + best record (in menu)
+
+    // Name formatter
+    std::string nameStr = std::string(levelData->name);
```
SuperV1234, line 31: `levelData->name` is already a string, no need to create another temporary one.

```suggestion
    std::string nameStr = levelData->name;
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -25,14 +25,28 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    nameStr[0] = std::toupper(nameStr[0]);
+
+    // Difficulty multipler formatter
+    std::string diffStr = std::to_string(HexagonGame::difficultyMult);
```
SuperV1234, line 35: `HexagonGame::difficultyMult` is redundant, just say `difficultyMult`.

```suggestion
    std::string diffStr = std::to_string(difficultyMult);
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -25,14 +25,28 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    size_t endPos = diffStr.find_last_not_of('0');
+    diffStr.erase(endPos + 1 + (int)(diffStr[endPos] == '.'), std::string::npos); // at least 1 dp
```
SuperV1234, line 37: I'm not sure what you're trying to achieve here, but it's very complicated. There must be a better way. What kind of formatting are you trying to get?

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -25,14 +25,28 @@ void HexagonGame::update(ssvu::FT mFT)
...
+
+    // Time formatter
+    std::string timeStr = std::to_string(floorf(status.getTimeSeconds() * 1000) / 1000); // 3 dp
```
SuperV1234, line 40: Just use `std::floor` instead of `floorf`.

```suggestion
    std::string timeStr = std::to_string(std::floor(status.getTimeSeconds() * 1000) / 1000); // 3 dp
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -25,14 +25,28 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    timeStr.erase(timeStr.find_first_of('.') + 4, std::string::npos);
```
SuperV1234, line 41: Again, this is not a convenient way of formatting... if you're trying to limit the number of floating point decimals, why not use something like `setprecision` or good ol'`printf`? https://stackoverflow.com/a/14432132/598696

src/SSVOpenHexagon/Core/HGUpdate.cpp
AlphaPromethium: This is so difficulty multipliers can be formatted to at least 1 decimal place.  For example, difficulty multipliers for Apeirogon could be displayed as [x0.35], [x0.5], [x0.75], [**x1.0**], [x1.25], [x1.4], [x1.6].

I was getting an issue where it was displaying as [**x1.**], and same for the time. I do feel it is a bit overcomplicated. Do ya have any suggestions on how I can improve this?

SuperV1234: Something like this?
https://gcc.godbolt.org/z/zWM8Ks

AlphaPromethium:
Outdated image. Difficulty _1, 2, 3, etc._ will now display as _[x1], [x2], [x3], etc._

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -21,18 +21,43 @@ using namespace hg::Utils;
 namespace hg
 {
 
+void nameFormat(std::string& name)
```
SuperV1234, line 24:
```suggestion
static void nameFormat(std::string& name)
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -21,18 +21,43 @@ using namespace hg::Utils;
...
+{
+    name[0] = std::toupper(name[0]);
+}
+
+[[nodiscard]] std::string diffFormat(float diff)
```
SuperV1234, line 29:
```suggestion
[[nodiscard]] static std::string diffFormat(float diff)
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -21,18 +21,43 @@ using namespace hg::Utils;
...
+{
+    char buf[255];
+    std::snprintf(buf, sizeof(buf), "%g", diff);
+    return buf;
+}
+
+[[nodiscard]] std::string timeFormat(float time)
```
SuperV1234, line 36:
```suggestion
[[nodiscard]] static std::string timeFormat(float time)
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -21,18 +21,43 @@ using namespace hg::Utils;
...
-    // TODO: refactor to avoid repetition, and truncate floating point number
-    // TODO: also show best record (here) and last run + best record (in menu)
-    steamManager.set_rich_presence_in_game(
-        levelData->name, status.getTimeSeconds());
+    // TODO: show best record (here) and last run + best record (in menu)
+
+    std::string nameStr = levelData->name;
+    nameFormat(nameStr);
+    std::string diffStr = diffFormat(difficultyMult);
+    std::string timeStr = timeFormat(status.getTimeSeconds());
```
SuperV1234, line 52:
```suggestion
    const std::string diffStr = diffFormat(difficultyMult);
    const std::string timeStr = timeFormat(status.getTimeSeconds());
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -21,18 +21,43 @@ using namespace hg::Utils;
...
-        levelData->name, status.getTimeSeconds());
+    // TODO: show best record (here) and last run + best record (in menu)
+
+    std::string nameStr = levelData->name;
+    nameFormat(nameStr);
+    const std::string diffStr = diffFormat(difficultyMult);
+    const std::string timeStr = timeFormat(status.getTimeSeconds());
+
+    // Presence formatter
+    std::string presenceStr = nameStr + " [x" + diffStr + "] - " + timeStr + "s";
```
SuperV1234, line 55:
```suggestion
    const std::string presenceStr = nameStr + " [x" + diffStr + "] - " + timeStr + "s";
```

PR #230, opened 2020-07-19
Bugfix/227 fix current time on steam not updating

### LINKED ISSUE #227: [Must Have] - Current time on Steam taking too long to update
When playing a level, the current time status on Steam takes way too long to update. I believe this is because requests are being made to the Steam API every frame, and needs to be completely reduced to every 1 or 2 seconds.

Steps to reproduce:
1. Start up the game.
2. Play any level.
3. Compare your current time in the game versus on Steam.

A/C:
1. Requests to the Steam API to update the current time should only be done every 1 second or 2, as opposed to every frame.
2. [Optional] Implemented the ability to programmatically change the time it takes to send a request to the Steam API.

AlphaPromethium:
Closes #227.

This fixes the current time on Steam not updating by adding an X second timer between the next call to the Steam API, where the value of X can be changed in the code.

The reason for the timer not updating on Steam is because the game is sending requests to the Steam API every frame, which is way too quick.

~~**_This PR should only be merged when #221 gets merged and when I resolve merge conflicts here, but I still want to get an approval for the time being!_**~~

### DIFF
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -59,7 +59,7 @@ public:
     CPlayer player;
     std::vector<CWall> walls;
     CCustomWallManager cwManager;
-    float timeUntilRichPresenceUpdate;
+    float timeUntilRichPresenceUpdate = 0.f;
 
 private:
     ssvs::Camera backgroundCamera{window,
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -59,6 +59,7 @@ public:
     CPlayer player;
     std::vector<CWall> walls;
     CCustomWallManager cwManager;
+    float timeUntilRichPresenceUpdate;
 
 private:
     ssvs::Camera backgroundCamera{window,
--- a/src/SSVOpenHexagon/Core/HGUpdate.cpp
+++ b/src/SSVOpenHexagon/Core/HGUpdate.cpp
@@ -60,13 +60,12 @@ void HexagonGame::update(ssvu::FT mFT)
     if (timeUntilRichPresenceUpdate <= 0.f)
     {
         steamManager.set_rich_presence_in_game(nameStr, diffStr, timeStr);
-        steamManager.run_callbacks();
-
         discordManager.set_rich_presence_in_game(presenceStr);
-        discordManager.run_callbacks();
-
         timeUntilRichPresenceUpdate = DELAY_TO_UPDATE;
     }
+
+    steamManager.run_callbacks();
+    discordManager.run_callbacks();
 
     hg::Joystick::update();
 
--- a/src/SSVOpenHexagon/Core/HGUpdate.cpp
+++ b/src/SSVOpenHexagon/Core/HGUpdate.cpp
@@ -51,15 +51,22 @@ void HexagonGame::update(ssvu::FT mFT)
     const std::string diffStr = diffFormat(difficultyMult);
     const std::string timeStr = timeFormat(status.getTimeSeconds());
 
+    constexpr float DELAY_TO_UPDATE = 5.f; // X seconds
+    timeUntilRichPresenceUpdate -= ssvu::getFTToSeconds(mFT);
+
     // Presence formatter
-    const std::string presenceStr =
-        nameStr + " [x" + diffStr + "] - " + timeStr + "s";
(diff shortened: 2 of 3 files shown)


src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -27,9 +27,18 @@ void HexagonGame::update(ssvu::FT mFT)
 
     // TODO: refactor to avoid repetition, and truncate floating point number
     // TODO: also show best record (here) and last run + best record (in menu)
-    steamManager.set_rich_presence_in_game(
-        levelData->name, status.getTimeSeconds());
-    steamManager.run_callbacks();
+
+    const int DELAY_TO_UPDATE = 5; // X seconds
```
SuperV1234, line 31:
```suggestion
    constexpr float DELAY_TO_UPDATE = 5; // X seconds
```

Make this `constexpr` (minor change, it means it's a compile-time constant) and `float` so it matches the type we're trying to compare against.

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -27,9 +27,18 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    static float timeTillUpdate = 0.0;
```
SuperV1234, line 32: This should be a data member of `HexagonGame` rather than a `static` variable. How about also calling it `timeUntilRichPresenceUpdate`?

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -27,9 +27,18 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    timeTillUpdate -= mFT / 60.0;
```
SuperV1234, line 33:
```suggestion
    timeTillUpdate -= ssvu::getFTToSeconds(mFT);
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -27,9 +27,18 @@ void HexagonGame::update(ssvu::FT mFT)
...
+
+    if (timeTillUpdate <= 0)
```
SuperV1234, line 35:
```suggestion
    if (timeTillUpdate <= 0.f)
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -27,9 +27,18 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    {
+        // only required for Steam
+        steamManager.set_rich_presence_in_game(levelData->name, status.getTimeSeconds());
+        steamManager.run_callbacks();
+        timeTillUpdate = DELAY_TO_UPDATE;
+    }
 
     discordManager.set_rich_presence_in_game(
```
SuperV1234, line 43: Doesn't Discord need the same treatment?

AlphaPromethium: Had a quick check by watching someone else play, and it does appear so. I will apply the same thing to discord and update the card/PR's description.

AlphaPromethium:
Discord now has the same treatment as it is also experiencing the same issue.

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -59,6 +59,7 @@ class HexagonGame
     CPlayer player;
     std::vector<CWall> walls;
     CCustomWallManager cwManager;
+    float timeUntilRichPresenceUpdate;
```
SuperV1234, line 62:
```suggestion
    float timeUntilRichPresenceUpdate = 0.f;
```

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -51,15 +51,22 @@ void HexagonGame::update(ssvu::FT mFT)
...
+
     // Presence formatter
-    const std::string presenceStr =
-        nameStr + " [x" + diffStr + "] - " + timeStr + "s";
+    const std::string presenceStr = nameStr + " [x" + diffStr + "] - " + timeStr + "s";
+
+    if (timeUntilRichPresenceUpdate <= 0.f)
+    {
+        steamManager.set_rich_presence_in_game(nameStr, diffStr, timeStr);
+        steamManager.run_callbacks();
```
SuperV1234, line 63: `run_callbacks()` should be executed per frame, both here and for `discordManager`

PR #231, opened 2020-07-19
Implement (more) deterministic RNG and fix rotation bug on Euclidean Plane Crash

### LINKED ISSUE #199: Rotation sometimes halts completely when level increments
This is a bug that is hard to reproduce, but it's bug that I'm coming across when playing a customly modified cube pack, where I keep track of stats to know the limits. 

Here's the best idea of how to reproduce it:
- Enable Dynamic FPS
- Try calling l_getRotationSpeed() on the function onIncrement() and write it to a variable.
- Track that variable. 

Depending on the level it is done on (Such as Apeirogon or Flattering Shape), for some reason the rotation speed just flat out goes to 0. For the life of me I don't know what the heck could be causing this, but if this can be figured out and solved that would be amazing.

SuperV1234:
Fixes #199, and keeps paving the way towards a replay system.


_RELEASE/config.json
```diff
@@ -29,7 +29,7 @@
 	"no_music" : false,
 	"no_rotation" : false,
 	"no_sound" : false,
-	"official" : true,
+	"official" : false,
```
Morxemplum, line 32: Please set this back to true. It's annoying when a build is off official mode by default.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -19,6 +19,14 @@ namespace hg
...
 void HexagonGame::initLua_Utils()
 {
+    addLuaFn("u_getRandomSeed", //
+        [this] { return rng.get_seed(); })
+        .doc(
+            "Obtain a random seed for use in Lua scripting. This function "
+            "should always be used to initialize `math.randomseed` before "
+            "generating random numbers, as it will guarantee determinism "
+            "across replays and enable score validation.");
+
```
Morxemplum, line 29: I'm confused. Will this make a new seed every time or will it get the generated seed for that attempt specifically?

_RELEASE/config.json
SuperV1234, line 32:
```suggestion
	"official" : true,

```

_RELEASE/config.json
```diff
@@ -30,6 +30,7 @@
 	"no_rotation" : false,
 	"no_sound" : false,
 	"official" : true,
+
```
SuperV1234, line 33:
```suggestion
```

src/SSVOpenHexagon/Core/HGScripting.cpp
SuperV1234: The former. Let me clarify the docs...

_RELEASE/Packs/cube/Scripts/utils.lua
```diff
@@ -1,19 +1,12 @@
...
 	local iterations = #t
 	local j
 	for i = iterations, 2, -1 do
 			j = math.random(i)
 			t[i], t[j] = t[j], t[i]
 	end
-	
+
 	return t
 end
```
Zly-u, line 11: Here i personally would go with:
```lua
function shuffle(t)
	for i = 1, #t do
		local j
		repeat j = math.random(#t) until i ~= j
		t[i], t[j] = t[j], t[i]
	end
end
```
So it prevents from swapping the same element from both i and j variables.
But it's not nessesary.

PR #238, opened 2020-07-22
WIP Replay System

SuperV1234:


AlphaPromethium:
Is it possible for this to be split into multiple cards? That way if someone else wants to help work on this too we can easily split up the work and track what's going on! Unless if ya wanna do it all on ya own that's fine too.

Morxemplum:
Looking great. Only problem is how do we access the replay?

Morxemplum:
Another thing that also bothers me and I know this is nitpicking, but why do some functions follow camel case with naming and others are followed with underscores? Looks pretty inconsistent to me.

SuperV1234:
> Another thing that also bothers me and I know this is nitpicking, but why do some functions follow camel case with naming and others are followed with underscores? Looks pretty inconsistent to me.

I've moved from the `CamelCase` style to `snake_case` during the years, and my rationale for mixing them here is that `snake_case`-styled components are easily recognizable as "recently written" and should therefore be of higher quality. End goal would be to gradually convert everything to `snake_case` when it reaches a good bar of quality.

SuperV1234:
> Looking great. Only problem is how do we access the replay?

This is still being worked on. At the moment it just starts automatically on restart. I'll think of an easy way of making this work

Morxemplum:
BUG: I tried playing Disc-o and sometimes when I level up I'm getting a Heptagon even though my max sides is 6? Is this a glitch with the game engine? I am also getting this on Massacre as well. Mind fixing this real quick?

AlphaPromethium:
This bug also occurs on other levels that do not have any side changing.

Morxemplum:
Another update on bug: This happens even on levels with side changing. It for some reason is raising the maximum sides to be 1 higher than the set value. This is a weird bug.

PR #239, opened 2020-07-22
Recoded collisions.

Zly-u:
New collision system that works for both curving and static walls.
Haven't been done for Custom walls though, comming soon`tm` but for now the standard method is okay.


include/SSVOpenHexagon/Components/CPlayer.hpp
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "SSVOpenHexagon/Utils/Ticker.hpp"
+#include "CWall.hpp"
```
SuperV1234, line 8: We don't need a physical dependency on `CWall.hpp` here, as we are only using `CWall` by reference (which does not require its definition). A [forward declaration](https://stackoverflow.com/questions/4757565/what-are-forward-declarations-in-c) would suffice. See line 18 in this file for an example. Reducing physical dependencies avoids dependency cycles and speeds up compilation time.

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -19,35 +19,29 @@
 #include <SFML/System/Vector2.hpp>
 #include <SFML/Graphics/Color.hpp>
 
-namespace hg
-{
+namespace hg {
```
SuperV1234, line 22: For the future, avoid changing the formatting of code. It makes the diffs bigger than they need to be. I would recommend using `clang-format` using the `.clang-format` style available in the repository root.

src/SSVOpenHexagon/Online/Online.cpp
```diff
@@ -516,7 +516,7 @@ void initializeValidators(HGAssets& mAssets)
 
         HG_LO_VERBOSE("hg::Online::initializeValidators")
             << "Finished initializing validators...\n";
-    */
+    //*/
```
SuperV1234, line 519:
```suggestion
    */
```

SuperV1234: Mistake?

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -266,6 +262,23 @@ void HexagonGame::update(ssvu::FT mFT)
         fpsWatcher.update();
     }
 }
+
+void HexagonGame::updateWalls(ssvu::FT mFT){
+    const auto playerPos{player.getPosition()};
```
SuperV1234, line 267: Unused?

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -136,17 +127,70 @@ void CPlayer::drawDeathEffect(HexagonGame& mHexagonGame)
...
+    if (!Config::getInvincible()) { dead = true; }
+    if (mEffect) { mHexagonGame.death(); }
+
+    ssvs::moveTowards(lastPos, ssvs::zeroVec2f, 5 * mHexagonGame.getSpeedMultDM());
+
+    pos = lastPos;
+}
+
+void CPlayer::push(HexagonGame& mHexagonGame, CWall& wall){
+    const auto curveData = wall.getCurve();
```
SuperV1234, line 159:
```suggestion
    const auto& curveData = wall.getCurve();
```

No need to copy here.

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -136,17 +127,70 @@ void CPlayer::drawDeathEffect(HexagonGame& mHexagonGame)
...
+    const int movement{mHexagonGame.getInputMovement()};
+
+    const float radius{mHexagonGame.getRadius()};
+
+    unsigned int maxAttempts;
+    if (curveDir != 0){
+        maxAttempts = std::abs(curveData.speed+speed*(movement*curveDir));
+    } else {
+        maxAttempts = speed;
+    }
```
SuperV1234, line 170:
```suggestion
    const unsigned int maxAttempts = curveDir != 0 
        ? std::abs(curveData.speed+speed*(movement*curveDir)) : speed;
```

SuperV1234: This is good because it allows us to mark `maxAttempts` as `const`, so we know it won't change in the rest of the function

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -136,17 +127,70 @@ void CPlayer::drawDeathEffect(HexagonGame& mHexagonGame)
...
+    unsigned int attempt = 0;
+    float pushDir;
+    if (curveDir != 0){
+        const int reverseDir = ((movement != 0) ? -movement : 1)*curveDir;
+        pushDir = ssvu::toRad(1.f)*curveDir*reverseDir;
+    } else {
+        pushDir = ssvu::toRad(1.f) * -movement;
+    }
```
SuperV1234, line 178: `const` would be nice here as well, but the logic is too complicated to put it in a ternary operator. It would be unreadable. We can use the IIFE pattern (immediately-invoked function expression):

```suggestion
    const float pushDir = [&]
    {
        if (curveDir == 0) {
            return ssvu::toRad(1.f) * -movement;
        }
       
        const int reverseDir = ((movement != 0) ? -movement : 1)*curveDir;
        return ssvu::toRad(1.f)*curveDir*reverseDir;
    }();
```

We are basically creating a lambda (anonymous function) and then immediately invoking it.

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -19,35 +19,29 @@
...
-    : startPos{mStartPos}, pos{startPos}, hue{0}, angle{0},
-      size{Config::getPlayerSize()}, speed{Config::getPlayerSpeed()},
-      focusSpeed{Config::getPlayerFocusSpeed()}, dead{false}, swapTimer{36.f},
-      swapBlinkTimer{5.f}, deadEffectTimer{80.f, false}
-{
+CPlayer::CPlayer(const sf::Vector2f &mStartPos) noexcept
+        : startPos{mStartPos}, pos{startPos}, hue{0}, angle{0},
+          size{Config::getPlayerSize()}, speed{Config::getPlayerSpeed()},
+          focusSpeed{Config::getPlayerFocusSpeed()}, dead{false}, swapTimer{36.f},
+          swapBlinkTimer{5.f}, deadEffectTimer{80.f, false} {
```
SuperV1234, line 30: We need to initialize `lastPos` to `startPos` here.

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -136,17 +127,70 @@ void CPlayer::drawDeathEffect(HexagonGame& mHexagonGame)
...
+        const int reverseDir = ((movement != 0) ? -movement : 1)*curveDir;
+        pushDir = ssvu::toRad(1.f)*curveDir*reverseDir;
+    } else {
+        pushDir = ssvu::toRad(1.f) * -movement;
+    }
+    while(wall.isOverlapping(pos)){
+        angle += pushDir;
+        pos = ssvs::getOrbitRad(startPos, angle, radius);
+
+        attempt++;
```
SuperV1234, line 183:
```suggestion
        ++attempt;
```

src/SSVOpenHexagon/Components/CPlayer.cpp
SuperV1234: We also need to initialize `lastAngle`.

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -188,9 +232,9 @@ void CPlayer::update(HexagonGame& mHexagonGame, ssvu::FT mFT)
 
     const sf::Vector2f tempPos{ssvs::getOrbitRad(startPos, angle, radius)};
     const sf::Vector2f pLeftCheck{
-        ssvs::getOrbitRad(tempPos, angle - ssvu::piHalf, 0.01f)};
+        ssvs::getOrbitRad(tempPos, angle - ssvu::piHalf, 1.01f)};
     const sf::Vector2f pRightCheck{
-        ssvs::getOrbitRad(tempPos, angle + ssvu::piHalf, 0.01f)};
+        ssvs::getOrbitRad(tempPos, angle + ssvu::piHalf, 1.01f)};
```
SuperV1234, line 237: Why was this changed?

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -136,17 +127,70 @@ void CPlayer::drawDeathEffect(HexagonGame& mHexagonGame)
...
 
+[[nodiscard]] sf::Vector2f CPlayer::getPosition() const {
+    return pos;
+}
+
+[[nodiscard]] sf::Vector2f CPlayer::getStartPosition() const {
+    return startPos;
+}
+
+void CPlayer::kill(HexagonGame &mHexagonGame, bool mEffect) {
```
SuperV1234, line 147: What's the point of `mEffect`? Is there anywhere when this should be called with `mEffect=false`? Also, the name is not very self-explanatory. What "effect" are we talking about?

src/SSVOpenHexagon/Components/CWall.cpp
```diff
@@ -62,15 +65,48 @@ void CWall::update(
...
+        if(moveCurve(mHexagonGame, mFT, vp, mCenterPos) && curve.speed != 0.f){
+            player.push(mHexagonGame, *this);
+        }
+    }
+
+    if(pointsOnCenter > 3)
+    {
+        killed = true;
+    }
+    */
```
SuperV1234, line 81: Let's remove this commented code block

src/SSVOpenHexagon/Components/CWall.cpp
```diff
@@ -62,15 +65,48 @@ void CWall::update(
...
+}
+
+bool CWall::moveTowardsCenter(HexagonGame& mHexagonGame, const sf::Vector2f& mCenterPos, ssvu::FT mFT){
```
SuperV1234, line 84: I think this should just return `void`, and we should explicitly check for collision directly in `HexagonGame::updateWalls`. This function has too many responsibilities

src/SSVOpenHexagon/Components/CWall.cpp
```diff
@@ -62,15 +65,48 @@ void CWall::update(
...
 
     if(pointsOnCenter > 3)
     {
         killed = true;
     }
+
+    return isOverlapping(mHexagonGame.player.getPosition());
+}
+
+bool CWall::moveCurve(HexagonGame& mHexagonGame, const sf::Vector2f& mCenterPos, ssvu::FT mFT){
```
SuperV1234, line 104: Ditto, should return `void` and we should check collision directly in the loop in `HexagonGame::updateWalls`

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -266,6 +262,23 @@ void HexagonGame::update(ssvu::FT mFT)
...
+
+    for(CWall& wall : walls) {
+        player.push(*this, wall);
+
+        if (wall.moveTowardsCenter(*this, centerPos, mFT)) {
```
SuperV1234, line 272: What I mean above is something like

```suggestion
        wall.moveTowardsCenter(*this, centerPos, mFT);
        if (wall.isOverlapping(player)) {
```

src/SSVOpenHexagon/Components/CPlayer.cpp
Zly-u: It's my IDE that does this, didn't found a function to disable this auto formatting unfortunately.

src/SSVOpenHexagon/Online/Online.cpp
Zly-u: a hacky way of abusing multi-line comments. Very handy for testing and such, when you put another slash at the beginning of it while the end of the comment already has double slash - it will uncomment the thing and you don't have to get rid of the ending of it - `*/`
```cpp
/*
void main(){ return 0; }
//*/
```
```cpp
//*
void main(){ return 0; }
//*/
```

Zly-u: Oh, but idk i didn't touch Online header though huh, well uh I thought that was my thing.

src/SSVOpenHexagon/Core/HGUpdate.cpp
Zly-u: alright, will change all of that soon.

src/SSVOpenHexagon/Components/CPlayer.cpp
Zly-u: idk, was testing stuff and didn't notice any difference so I guess I left it for more precise collision detection...? Eh, idk honestly but this part will need to be changed actually dues to the custom walls being a complete whacks to properly calculate for collisions, so we need to leave it as just:

```cpp
if(wall.isOverlapping(pos)){
    kill(...);
}
```

src/SSVOpenHexagon/Components/CPlayer.cpp
Zly-u: wanted to use that function to replace standard methods of killing s player to simplify it.
And by the effect it means the death effect itself... Prob had to name it like this.

Zly-u: also yes there are places where it supposed to be called with `mEffect=false` like for example while `tutorial mode` is on iirc.

Zly-u: also yes there are places where it supposed to be called with `mEffect=false` like for example while `tutorial mode` is on iirc.

Issue #150, opened 2017-09-20
OSX build?


CatsAreFluffy:

SuperV1234:
I don't own a Mac, but I see no particular reason this wouldn't build on OSX. You are welcome to try building if you own a Mac - let me know how it goes and if you have any question.

CatsAreFluffy:
Well, I'm kinda lazy, so i just used Wine :P

scemino:
I'm also interested in OSX build.
Here is the full build file on my mac:
[build.txt.zip](https://github.com/SuperV1234/SSVOpenHexagon/files/4944363/build.txt.zip)

Morxemplum:
We do have a few Mac users in the Open Hexagon community so if you manage to get a native build working for MacOS, that'd be wonderful! :)

AlphaPromethium:
I think we should at least try to get this working before full release in August. Would be epic!

Morxemplum:
Only real issue is that none of us own a Mac so it's really hard for us to even try to build the game for Mac, considering how picky Apple is when it comes to who has their OS

AlphaPromethium:
We could try renting a MacInCloud for a month?

Morxemplum:
Never heard of it. But I was thinking of getting a family Macbook for my family so at least there's a family computer that my parents can use for their stuff. This is an interesting topic though.

scemino:
I can help you, if you want.

Morxemplum:
All contributions are welcome! :)

scemino:
I have still an error:
`LuaWrapper.hpp:1318:19: error: expected body of lambda expression
        return [&]<int... Is>(std::integer_sequence<int, Is...>)`
it should be valid with C++ 20 enabled in CMakeLists.txt -> set(CMAKE_CXX_STANDARD "20")
If someone knows what to do...

scemino:
So I updated clang to version 11.0.3, now it's different:
```
[build] clang: error: unable to execute command: Segmentation fault: 11
[build] clang: error: clang frontend command failed due to signal (use -v to see invocation)
[build] Apple clang version 11.0.3 (clang-1103.0.32.62)
[build] Target: x86_64-apple-darwin19.4.0
```
I don't know if C++20 is stable in clang ;)

SuperV1234:
@scemino: is there any way you could use GCC 9.2.x or higher on Mac OS?

scemino:
[...]

Issue #202, opened 2020-07-13
[Must Have] - Setup Unit Tests


AlphaPromethium:
Should be simple to understand—when someone implements a feature, fixes a bug, etc., every issue should be tested with a unit test so that it's proven to work and it doesn't accidentally break something else. No one should expect everything to work 100% of the time.

Running unit tests should be done in its own directory, _"tests"_. For the game's UI, unit tests should not be testing elements on the UI, but rather the code called by an element.

File structure may be dependent on the unit testing framework used, but ideally something like this should be used:
```
tests
|_ SSVOpenHexagon
   |_ Components
   |  |_ CCustomWallTests.cpp
   |  |_ CPlayerTests.cpp
   |  |_ CWallTests.cpp
   |
   |_ Core
   |  |_ CCustomWallManagerTests.cpp
   |  |_ DiscordTests.cpp
   |  |_ HGGraphicsTests.cpp
   |  |_ HGScriptingTests.cpp
   |  ...
   ...
```

A/C:
1. A unit test framework is used and "tests" directory is setup.
2. [Optional] For every recent issue worked on, at least 1 unit test should be written to test that it works as expected. Could put this as a card of its own.
3. [Optional] The game's core gameplay code should be tested. Could put this as a card of its own.

_Once this issue is complete, every new issue being worked on should have at least 1 unit test, and PR reviewers will be able to inspect the test. Optionally, we could have a new column in the "Projects" tab named "Testing" for collaborators to test the new feature/bugfix/etc. and also inspect the created unit test(s) for that issue._

**_Priority: Must Have_.**

AlphaPromethium:
Feel free to leave any unit testing framework suggestions here! I had a quick look and Google Test (GTest) may be suitable.

Morxemplum:
I agree that regression testing is an absolute must for any decent program, and I don't think Open Hexagon is any exception. A test suite for Open Hexagon as a whole would be very nice to have to not only help patch bags, but heavily lower the odds of any new bugs appearing.

SuperV1234:
I've always wanted to try [`doctest`](https://github.com/onqtam/doctest). It seems modern and super lightweight. This is a good opportunity.

AlphaPromethium:
Dragging this back into todo for the time being as I am not able to get enough time to work on it. If anyone else would like to pick this up and implementing doctest that is fine by me! Just be sure to pull my branch to get my current progress! :>

SuperV1234:
[...]

PR #240, opened 2020-07-23
Add CI builds for Linux

InfoTeddy:
This PR adds automatic CI builds for Linux. Every time a push or pull request is made, GitHub Actions will spin up an Ubuntu container that automatically checks out the repository, downloads dependencies, builds the game, and uploads the result. Anyone with a GitHub account can go to the Actions page, click the latest run, and download said build.

These CI builds serve two purposes: (1) to check that changes made to the codebase will compile correctly, and (2) to provide convenient builds so players don't have to clone and compile the game for themselves. (However, it's still a bit annoying, because you have to make a GitHub account and log in, but it's better than nothing.)

I would have liked to add CI builds for Windows and macOS, but every approach I tried for either of those two operating systems failed to work. I have spent countless hours attempting to get Windows builds working, but every attempt I just ran into annoying roadblocks that weren't easy to get around. The problem seems to be the fact that Open Hexagon uses so much new and exotic C++20 constructs, which requires GCC 9 or later, and which the corresponding MinGW 9 or later is hard to obtain. Not helping the case is that MSVC is simply not supported by Open Hexagon, so I can't go down that route either. And as for macOS, well, #150 has already been filed, and it seems to me that Open Hexagon depends on X11/XRandR stuff, which is hard to support on macOS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,40 @@
+name: CI
+
+# Trigger this workflow on push or pull request
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Build (ubuntu-20.04)
+
+    runs-on: ubuntu-20.04
+
+    env:
+      CXXFLAGS: -I/usr/local/include/SFML -std=c++2a
+      LDFLAGS: -L/usr/local/lib
+
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: recursive
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y ninja-build libsfml-dev libx11-dev libxrandr-dev libudev-dev libopenal-dev libvorbis-dev libflac-dev
+
+    - name: CMake configure
+      run: |
+        cd build
+        cmake -GNinja ..
+    - name: Build
+      run: ninja -C build
+
+    - name: Copy artifacts
+      run: cp build/SSVOpenHexagon build/OHWorkshopUploader _RELEASE
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v1
+      with:
+        name: OpenHexagon-Linux
+        path: _RELEASE
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     - name: CMake configure
       run: |
         cd build
-        cmake -GNinja ..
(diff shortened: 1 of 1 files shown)


InfoTeddy:
As a demonstration, you can [take a look at the CI run for this very branch](https://github.com/InfoTeddy/SSVOpenHexagon/actions/runs/179421877). And if you have a GitHub account and are logged in, then you can [download the Linux build](https://github.com/InfoTeddy/SSVOpenHexagon/suites/953609296/artifacts/11921077) that was automatically generated.

InfoTeddy:
Also, once a build has been compiled, we'd be able to run tests on it (#202).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
+name: CI
+
+# Trigger this workflow on push or pull request
+on: [push, pull_request]
```
AlphaPromethium, line 4: Will one need to be triggered when a pull request is created? Ya need to push your changes to a branch anyway before one can be made anyway.

AlphaPromethium: _(Ignore review made. It was a misclick.)_

AlphaPromethium:
This is really nice! We can also integrate the unit tests into the CI as well once that gets done!

.github/workflows/ci.yml
InfoTeddy: Yes, one gets triggered upon push anyway, but that run doesn't show up on the pull request page. The pull request run is to make it easy for reviewers to see if the PR succeeds or fails to compile.

AlphaPromethium: Ah, ok. Fair enough.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y ninja-build libsfml-dev libx11-dev libxrandr-dev libudev-dev libopenal-dev libvorbis-dev libflac-dev
+
+    - name: CMake configure
+      run: |
+        cd build
+        cmake -GNinja ..
```
SuperV1234, line 29:
```suggestion
        cmake -GNinja -DCMAKE_BUILD_TYPE=RELEASE ..
```

SuperV1234:
Really great stuff, @InfoTeddy - thanks! I apologize for the painful experience trying to set Windows CI up - I will look around and see if there's anything that can be done to get that working. This is a fantastic start, though.

PR #242, opened 2020-07-24
Allow styles to support negative values for the `"Pulse" field

### LINKED ISSUE #241: Allow color pulses to properly support negative values in Styles
This was hiding in the back of my head and it didn't come to light until recently. I've noticed with the pulse field that it doesn't properly support negative values. Why is this a problem? This limits the possibilities of styles that can be made. 

A good example to demonstrate: Start off with a color of (0, 255, 0, 255), which is a fully bright green. Now pulse that by (255, -255, 255, 0) to get (255, 0, 255, 255), which is magenta. 

However, if you try to do this using Open Hexagon styles, you will actually pulse to white instead of magenta, which easily tells me that negative values do not have proper support. There is a demonstration of negative pulses in the Pointless style, but I don't feel like these are negative pulses being used properly (and as far as I know, I don't see a noticable difference if you remove the negative value).

This may either be easy to implement or very difficult. But all I know is that it is important to support negative values in pulsing to increase the number of possibilities that can be created using pulses.

Morxemplum:
Closes #241.

Had to make my own struct and all that to accomplish this. I wouldn't have been able to do this without a bit of help from @SuperV1234  and @MultHub, because I was struggling to figure out how the JSON Parser worked when parsing arrays. 

Hopefully people can now make all new types of styles with this restriction removed.

### DIFF
--- a/include/SSVOpenHexagon/Data/ColorData.hpp
+++ b/include/SSVOpenHexagon/Data/ColorData.hpp
@@ -15,11 +15,6 @@ struct PulseColor {
     int g;
     int b;
     int a;
-
-    PulseColor() = default;
-
-    PulseColor(const int red, const int green, const int blue, const int alpha)
-        : r{red}, g{green}, b{blue}, a{alpha} {};
 };
 
 namespace hg 
--- a/include/SSVOpenHexagon/Data/ColorData.hpp
+++ b/include/SSVOpenHexagon/Data/ColorData.hpp
@@ -44,7 +44,7 @@ struct ColorData
 
     ColorData(const bool mMain, const bool mDynamic, const bool mDynamicOffset,
         const float mDynamicDarkness, const float mHueShift,
-        const float mOffset, sf::Color mColor, PulseColor mPulse)
+        const float mOffset, sf::Color mColor, const PulseColor& mPulse)
         : main{mMain}, dynamic{mDynamic}, dynamicOffset{mDynamicOffset},
           dynamicDarkness{mDynamicDarkness}, hueShift{mHueShift},
           offset{mOffset}, color{mColor}, pulse{mPulse}
--- a/include/SSVOpenHexagon/Data/ColorData.hpp
+++ b/include/SSVOpenHexagon/Data/ColorData.hpp
@@ -9,11 +9,30 @@
 
 #include <SFML/Graphics/Color.hpp>
 
+
+struct PulseColor {
+    int r;
+    int g;
+    int b;
+    int a;
+
+    PulseColor() = default;
+
+    PulseColor(const int red, const int green, const int blue, const int alpha)
+        : r{red}, g{green}, b{blue}, a{alpha} {};
+};
+
+namespace hg 
+{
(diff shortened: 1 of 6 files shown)


Morxemplum:
There's also a new experimental level you can test out! :) Might reuse it for an actual level? 🤔

src/SSVOpenHexagon/Data/ColorData.cpp
```diff
@@ -0,0 +1,22 @@
...
+namespace hg 
+{
+	[[nodiscard]] PulseColor pulse_from_json(const ssvuj::Obj& root) noexcept
+	{
+		std::vector<int> rawPulseData = ssvuj::getExtr<std::vector<int>>(root, "pulse", {0, 0, 0, 255});
+		if (rawPulseData.size() == 4)
+		{
+			return {rawPulseData[0], rawPulseData[1], rawPulseData[2], rawPulseData[3]};
+		}
+		return {0, 0, 0, 255};
```
SuperV1234, line 20:
```suggestion
        if(!ssvuj::hasObj(root, "pulse"))
        {
             return {0, 0, 0, 255};
        }
        
        const auto& pulseObj = ssvuj::getObj(root, "pulse");

        return {
            ssvuj::getExtr<int>(pulseObj, 0),
            ssvuj::getExtr<int>(pulseObj, 1),
            ssvuj::getExtr<int>(pulseObj, 2),
            ssvuj::getExtr<int>(pulseObj, 3)
        };
```

The above should work and avoid any unnecessary dynamic allocation.

include/SSVOpenHexagon/Data/ColorData.hpp
```diff
@@ -26,11 +45,11 @@ struct ColorData
           hueShift{ssvuj::getExtr<float>(mRoot, "hue_shift", 0.f)},
           offset{ssvuj::getExtr<float>(mRoot, "offset", 0.f)},
           color{ssvuj::getExtr<sf::Color>(mRoot, "value", sf::Color::White)},
-          pulse{ssvuj::getExtr<sf::Color>(mRoot, "pulse", sf::Color::White)} {};
+          pulse{hg::pulse_from_json(mRoot)} {};
 
     ColorData(const bool mMain, const bool mDynamic, const bool mDynamicOffset,
         const float mDynamicDarkness, const float mHueShift,
-        const float mOffset, sf::Color mColor, sf::Color mPulse)
+        const float mOffset, sf::Color mColor, PulseColor mPulse)
```
SuperV1234, line 52:
```suggestion
        const float mOffset, sf::Color mColor, const PulseColor& mPulse)
```

include/SSVOpenHexagon/Data/ColorData.hpp
```diff
@@ -9,11 +9,30 @@
...
+struct PulseColor {
+    int r;
+    int g;
+    int b;
+    int a;
+
+    PulseColor() = default;
+
+    PulseColor(const int red, const int green, const int blue, const int alpha)
+        : r{red}, g{green}, b{blue}, a{alpha} {};
```
SuperV1234, line 22:
```suggestion
```

We don't need these

PR #246, opened 2020-07-26
[URGENT] Destroy Lua Libraries and Functions That Can Be Used Maliciously

### LINKED ISSUE #237: [Must Have] - Remove (or heavily restrict) the "io" and "os" libraries in Lua
For anybody who has a deep knowledge of Lua, you know that this is an issue that must be taken seriously. 

The "io" and "os" libraries are very powerful libraries in Lua. Some of the functions are harmless, but there are others that are capable of doing serious damage. One of the best examples being "os.execute", which can allow Lua scripts to run commands to the terminal of a given operating system. This can be used to perform malicious actions, such as shutting down a computer or deleting entire hard drives (both of these examples have actually been done before in Open Hexagon history).

The "io" library is capable of doing the same damage, such as overwriting critical files, jamming other packs to be dysfunctional, or write malicious files in the first place. We can either remove this library completely or restrict what file types it can write and where it can write them. 

I think it's important that for the upcoming release, it's important that this issue must be handled, or people can start using Open Hexagon as a weapon against people they don't like. It has been done before, and I don't want to see future cases of this happening.

### LINKED ISSUE #244: Game crashes when a Lua error happens on "onUnload"
When a Lua error happens with onUnload, an infinite feedback loop happens and as a result crashes the game. I've figured out the problem and can fix this easily.

Morxemplum:
Closes #237 and fixes #244 

Things that will be destroyed:
- The entire "os" library
- The entire "io" library
- The "math.randomseed" function

This is an URGENT pull request that is needed, because this will fix critical security issues of Open Hexagon. This stops bad faithed people from being able to use Lua code to write malware and be able to harm computers of Open Hexagon players. 

If you can, please try and push this fix into the main branch of Open Hexagon for Steam, because this is something super serious and we want to apply this fix quickly before someone writes malware for the game.

### DIFF
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -161,7 +161,8 @@ private:
 
     Utils::LuaMetadata luaMetadata;
 
-    // LUA-related methods
+    // Lua related methods
+    void destroyMaliciousFunctions();
     void initLua_Utils();
     void initLua_Messages();
     void initLua_MainTimeline();
--- a/src/SSVOpenHexagon/Core/HGScripting.cpp
+++ b/src/SSVOpenHexagon/Core/HGScripting.cpp
@@ -31,9 +31,22 @@ void HexagonGame::destroyMaliciousFunctions()
     // contents, which can be used maliciously.
     lua.clearVariable("io");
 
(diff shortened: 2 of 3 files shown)


Morxemplum:
Some new functions will be destroyed: The "debug" library, package.loadlib, and package.searchpath.

PR #252, opened 2020-07-29
Add `onInput` Lua callback

SuperV1234:
> * **`bool onInput(float mFrameTime, int mMovement, bool mFocus, bool mSwap)`**: A core function that is called every frame throughout the level, with the argument `mFrameTime` being a float representing the time elapsed for a single frame. The `mMovement` argument represents the movement input provided by the player (`-1` for counter-clockwise rotation, `1` for clockwise rotation, `0` otherwise). The `mFocus` and `mSwap` boolean arguments are set to `true` if the focus or swap inputs are being held down, respectively. The return value of the function dictates whether the inputs should be used by the engine to move the player, or not. Returning `false` will prevent the player from moving, focusing, or swapping --- regardless of the user input. Returning `false` can be useful for levels that completely change the movement mechanics, or that restrict some certain actions. Use this function wisely --- the same considerations as per `onUpdate` apply.


Morxemplum:
While **onInput** wasn't the name that I was looking for and for it to run every frame, at least this addresses a lot of the concerns that I have made in the issue that I have linked above. I like it and I think it's good for basic input.

include/SSVOpenHexagon/Utils/Utils.hpp
```diff
@@ -83,6 +84,13 @@ inline void runLuaFile(Lua::LuaContext& mLua, const std::string& mFileName)
     }
 }
 
+struct Nothing
+{
+};
```
Morxemplum, line 89: What's the purpose of the Nothing struct?

SuperV1234: It's a workaround for C++ silliness, as `void` is not a regular type and you cannot create instances of it. `runLuaFunctionIfExists` is supposed to return a `std::optional<T>` where `T` is the return value of the Lua function. This is great if the Lua function actually returns something, but if it returns `void`, the type `std::optional<void>` is invalid. Therefore I perform a transformation from `void` to `Nothing` during the function, so that I can homogeneously handle `void` as any other type, and stick it inside a `std::optional`.

PR #264, opened 2020-08-01
Added a path to script folder for `require` to work properly.

Zly-u:
Still in WIP in some way, will try to figure out a better way to do it.
Meanwhile, you can suggest your ideas for what method would fit this better.


Morxemplum:
Don't link an issue if you're using a checklist. I wish GitHub was that sophisticated and could allow you to check off tasks using pull requests.

Zly-u:
dunno how that is a problem since you can link several issues but okay i guess..?

Morxemplum:
Because if you close the issue then it's assumed you've completed the entire check list (which I don't think is the case here). You can only link issues, not the tasks contained within them.

Zly-u:
Uuuhhhh what. K whatever that just makes no sense to me.

Morxemplum:
It doesn't, and again, I wish GitHub had this as a feature, but unfortunately we don't.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -6,9 +6,6 @@
 #include "SSVOpenHexagon/Global/Assets.hpp"
 #include "SSVOpenHexagon/Utils/Utils.hpp"
 #include "SSVOpenHexagon/Core/HexagonGame.hpp"
-#include "SSVOpenHexagon/Components/CWall.hpp"
-#include "SSVOpenHexagon/Components/CCustomWallHandle.hpp"
-#include "SSVOpenHexagon/Components/CCustomWall.hpp"
```
Morxemplum, line 11: Why are you removing this? There are custom wall functions defined here that need these headers.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -1216,23 +1213,31 @@ void HexagonGame::initLua()
...
-            "Obtain the current random seed, automatically generated at the "
-            "beginning of the level. `math.randomseed` is automatically "
-            "initialized with the result of this function at the beginning of "
-            "a level.");
+             [this]{ return rng.seed(); })
+            .doc(
+                    "Obtain the current random seed, automatically generated at the "
+                    "beginning of the level. `math.randomseed` is automatically "
+                    "initialized with the result of this function at the beginning of "
+                    "a level.");
```
Morxemplum, line 1221: Be sure to stick to the ``.clang-format`` that Vee has defined for us.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -1216,23 +1213,31 @@ void HexagonGame::initLua()
...
+    try{
         lua.executeCode("math.randomseed(u_getAttemptRandomSeed())");
     }
-    catch(...)
-    {
+    catch(...){
         ssvu::lo("HexagonGame::initLua")
-            << "Failure to initialize Lua random generator seed\n";
+                << "Failure to initialize Lua random generator seed\n";
     }
```
Morxemplum, line 1241: Same thing applies here.

Morxemplum (changes requested):
Uhhhhhh. There's some good changes, but some others seem pretty sketchy, I'd like an explanation as to why some of those changes were made.

src/SSVOpenHexagon/Core/HGScripting.cpp
Zly-u: they are not even used in the include.

src/SSVOpenHexagon/Core/HGScripting.cpp
Zly-u: can you go away with it when the thing is god damn raw?

PR #272, opened 2020-08-10
Replace all HTTP occurrences with HTTPS

Morxemplum:
Small action, but **affects every file**. I essentially just went into Studio and just replaced every occurrence of "http://" with "https://".

This change should do virtually nothing. It just freshens up the documentation a bit to meet more modern standards.


SuperV1234 (approved):
This should probably be fine.

PR #273, opened 2020-08-12
Get Camera Shake to accept float values

Morxemplum:
I consider this more of a bug fix than a feature. Previously setting camera shake wouldn't accept float values. Now it can. Pretty simple fix.


SuperV1234 (approved):
LGTM

PR #276, opened 2020-08-12
Menu bind option

IvoryDuke:
NOTE! In order for this feature to be implemented the pull requests I made to the SSVStart, SSVMenuSystem libraries need to be merged.
 
Brief: Added menu items that allow in game rebinds.

Detailed:
- keyboard and mouse, and joystick binds are supported. 2 keys can be assigned to the keyboard binds, joystick binds only take one button;
- upon loading the game the config.json where the binds are stored is "sanitized". Unknown keys/buttons are replaced by an empty bind user can customize in game. If a key is bound to multiple functions the latter binds will be removed;
- to change a bind user must press Return, or the Select joystick button. The possibility to enter a bind will be marked by a "_" a the end of the item name;
- the player's pressed key will be ignored if a key assigned to a hardcoded function has been pressed;
- if user presses a key/button has already been assigned to another function the previous bind is removed and the key is reassigned to the one player is currently customizing;
- while the menu item is waiting for a player inputted key all other binded functions are disabled. Ex. pressing down will not move cursor to the next item in the menu;
- binds can be removed by pressing backspace. Some functions have been assigned extra hardcoded keys to prevent user from screwing itself over. For ex. even if player has no Up or Down keys bound the arrows will still allow it to navigate the menus. The hardcoded binds are of course non customizable, but user can assign extra keys _on top_ of those;
- when binds are assigned/removed the corresponding actions are updated. Aka, if I add "A" to rotate couterclockwise I can immeditely use it in game;
- the edited binds are stored in the .json file when exiting the game.

Short video demo
https://youtu.be/PTZ6slV0z9U


IvoryDuke:
As expected the build failed since I could not push the SSV edits, please make sure to check the attached zip file to compile successfully.

InfoTeddy:
If you changed SSV libs then you should make the relevant pull requests to those repositories first instead of having them be a separate attachment.

IvoryDuke:
I asked suggestions on how to go about it in the Discord and was just told to not commit.
I got no better answer until now.
Where should I make the pull request for those?

InfoTeddy:
Well, since it looks like you modified SSVMenuSystem, SSVStart, and SSVUtils, you should go to [SuperV1234/SSVMenuSystem](https://github.com/SuperV1234/SSVMenuSystem), [SuperV1234/SSVStart](https://github.com/SuperV1234/SSVStart), and [SuperV1234/SSVUtils](https://github.com/SuperV1234/SSVUtils).

IvoryDuke:
Alright thank you, I was just about to type that I had found them.
It will take me a sec but I'll see to it in a moment

IvoryDuke:
@InfoTeddy I have updated those libraries accordingly.
As mentioned in the comment on SSVStart I aknowledge perhaps that Open Hexagon specific enumerators does not fit in the context of that library.
I'll see to move it in a more appropriate file in the next few days.
Please forgive my noobish mistake and thank you very much for taking time to point me in the right direction.

IvoryDuke:
A video showing the changes in the last update I pushed
https://youtu.be/9_nQvnZZdhU

Joystick support required handling the bindings in a slightly different way since they interface with the menus and game in a different manner compared to keyboard and mouse but it works in the same way for users.
A change that was brough over to keyboard/mouse as well is that when you try assigning a key/button that is already bound to another function this key is first unbound from that function and reassigned to the new one.
It did increase the number of functions in Config.hpp and Config.cpp, but it is lighter on the menus and personally I like this solution better.

_RELEASE/config.json
```diff
@@ -47,12 +58,13 @@
 	"sound_volume" : 75.0,
 	"t_down" : 
 	[
-		[ "kDown" ],
-		[ "kS" ]
+		[ "kS" ],
+		[ "" ]
 	],
```
SuperV1234, line 63: Should these changes be committed at all?

misc/default_config.json
```diff
@@ -40,65 +41,76 @@
...
+		[ "" ]
 	],
-	"t_restart" :
+	"t_replay" : 
 	[
-		[ "kReturn" ],
-		[ "bMiddle" ]
+		[ "" ],
+		[ "" ]
 	],
```
SuperV1234, line 72: Ditto, should these changes be committed? Don't we want kReturn and bMiddle to be defaults?

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -3,20 +3,9 @@
...
-#include "SSVOpenHexagon/Core/Steam.hpp"
-#include "SSVOpenHexagon/Core/Discord.hpp"
-#include "SSVOpenHexagon/Online/Online.hpp"
-#include "SSVOpenHexagon/Utils/Utils.hpp"
-#include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
-
-#include <SSVStart/Utils/Vector2.hpp>
-#include <SSVStart/SoundPlayer/SoundPlayer.hpp>
-
-#include <SSVUtils/Core/Common/Frametime.hpp>
```
SuperV1234, line 19: Please keep the includes

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -122,20 +113,27 @@ HexagonGame::HexagonGame(Steam::steam_manager& mSteamManager,
...
     game.addInput(
         Config::getTriggerForceRestart(),
         [this](ssvu::FT /*unused*/) {
             status.mustStateChange = StateChange::MustRestart;
         },
-        ssvs::Input::Type::Once);
+        ssvs::Input::Type::Once, Tid::ForceRestart);
     game.addInput(
-        Config::getTriggerRestart(),
+        {{sf::Keyboard::Key::R}},                                                           //hardcoded
```
SuperV1234, line 136: Why do we want to hardcode `R`?

src/SSVOpenHexagon/Core/Joystick.cpp
```diff
@@ -85,6 +108,17 @@ struct JoystickState
     return res;
 }
 
+void setJoystickBind(unsigned int button, Jid buttonID)
+{
+    auto& s = getJoystickState();
+    s.joystickInputs[int(buttonID)] = button;
```
SuperV1234, line 114:
```suggestion
    getJoystickState().joystickInputs[int(buttonID)] = button;
```

src/SSVOpenHexagon/Core/Joystick.cpp
```diff
@@ -85,6 +108,17 @@ struct JoystickState
...
+}
+void unbindJoystickButton(unsigned int buttonID)
+{
+    auto& s = getJoystickState();
+    s.joystickInputs[buttonID] = 33;
```
SuperV1234, line 119:
```suggestion
    getJoystickState().joystickInputs[buttonID] = 33;
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -6,18 +6,6 @@
...
-#include "SSVOpenHexagon/Online/Online.hpp"
-#include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
-#include "SSVOpenHexagon/SSVUtilsJson/SSVUtilsJson.hpp"
-
-#include <SSVStart/Input/Input.hpp>
-#include <SSVStart/Utils/Vector2.hpp>
-
-#include <SSVMenuSystem/SSVMenuSystem.hpp>
-
-#include <SSVUtils/Core/Common/Frametime.hpp>
```
SuperV1234, line 20: Please keep the includes

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -68,6 +56,88 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
                 downAction();
             }
         };
+    game.onEvent(Event::EventType::KeyReleased) +=
+        [this](const Event& mEvent)
+        {
+            if(!noActions) return;
```
SuperV1234, line 62: The name `noActions` is not immediately clear -- please rename the variable or add a comment. I'm not sure what it means

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -559,19 +749,26 @@ void MenuGame::initInput()
...
-        [this](ssvu::FT /*unused*/) { leftAction(); }, t::Once);
+        [this](ssvu::FT /*unused*/) { leftAction(); }, t::Once, Tid::RotateCCW);
     game.addInput(
         Config::getTriggerRotateCW(),
-        [this](ssvu::FT /*unused*/) { rightAction(); }, t::Once);
+        [this](ssvu::FT /*unused*/) { rightAction(); }, t::Once,
+        Tid::RotateCW);
     game.addInput(
-        Config::getTriggerUp(), [this](ssvu::FT /*unused*/) { upAction(); },
+        {{k::Up}}, [this](ssvu::FT /*unused*/) { upAction(); }, //hardcoded
```
SuperV1234, line 758: Ditto, why should this be hardcoded?

src/SSVOpenHexagon/Global/Config.cpp
```diff
@@ -5,15 +5,9 @@
...
-#include "SSVOpenHexagon/Online/Online.hpp"
-#include "SSVOpenHexagon/SSVUtilsJson/SSVUtilsJson.hpp"
-
-#include <SSVStart/Input/Input.hpp>
-#include <SSVStart/GameSystem/GameWindow.hpp>
+#include "SSVOpenHexagon/Core/Joystick.hpp"
 
 #include <iostream>
-#include <fstream>
-#include <memory>
```
SuperV1234, line 16: Please keep the includes

SuperV1234 (changes requested):
Very promising and hard work -- thanks! This is just a preliminary review, as I need to look at the changes in the extlibs before I can go deeper on this one. A few things can be addressed now, though.

src/SSVOpenHexagon/Core/HexagonGame.cpp
IvoryDuke: I made it hardcoded because it is the key that player is told to be pressed to restart after it dies

misc/default_config.json
IvoryDuke: That must have been an oversight on my end, I will recheck the .json

src/SSVOpenHexagon/Core/MenuGame.cpp
IvoryDuke: The idea is that no matter how much the user fools around with the binds, it should never be able to "screw itself over".
For example, if we were to not have hardcoded Up and Down binds, and user were to unassign all keys from those action, it would practically be unable to navigate the menues anymore.
Therefore, unless this explanation has not been convincing, I strongly feel we should at least keep the most essential hardcoded inputs I created.

src/SSVOpenHexagon/Core/MenuGame.cpp
IvoryDuke: I named it after the same variable that is part of the DialogBox PR.
When 1 or higher it's a signal to MenuGame that all actions like upAction(), downAction() etc. are temporarely disabled since user should be able to assign a key to a bind without any of those actions be triggered in the process, and the key presses can be used by the function below to be assigned to the binds.

If it's not clearn enough I will change the name and also change the name in the other PR for consistency.

SuperV1234 (approved):
LGTM. Will clean up a bit. Thanks!

SuperV1234 (approved):
LGTM

PR #278, opened 2020-08-14
Allow Pack Developers to Override the Timer for Alternative Scoring

### LINKED ISSUE #274: Introduce a method of Alternative Scoring
By default, Open Hexagon uses a timer for how long the player survives to score an attempt. This will always be the default method of scoring in Open Hexagon. However, from seeing some of the creativity in some of the player packs, some levels don't suit very well when scored by the amount of time they've survived in an attempt. 

What I plan on doing is being able to override the default scoring value and be able to bind it to a custom Lua variable that the user defines, and the variable that has been bound will replace the timer as a way of scoring. This will also be bound to personal bests and will also be prevalent with the level's leaderboard. 

This can actually make a lot more level ideas be possible by introducing a way to customize how levels can be scored. I think this is a "Should Have"

Morxemplum:
Closes #274 

This pull request will add a Lua function that allows pack developers to override the traditional scoring used in Open Hexagon, which is based on how long someone survives. The function is called ``l_overrideScore``, and takes in a string value that is the name of a Lua variable. Using that information it bases the score off the value of the variable.

This function ONLY accepts number values. If a Lua variable of any other type is used to override the score, a Lua error will be made and kick players back to the menu. We don't support strings at the moment because they can be difficult to sort and we certainly don't want to use booleans or tables for the scoring.

This function is very useful for levels that can't be scored by time survived, and need to resort to different methods of scoring.

### DIFF
--- a/src/SSVOpenHexagon/Core/HGGraphics.cpp
+++ b/src/SSVOpenHexagon/Core/HGGraphics.cpp
@@ -290,13 +290,23 @@ void HexagonGame::updateText()
     os.flush();
 
     // Set in game timer text
-    if(status.started)
+    if(!levelStatus.scoreOverridden)
     {
-        timeText.setString(formatTime(status.getTimeSeconds()));
+        // By default, use the timer for scoring
+        if(status.started)
+        {
+            timeText.setString(formatTime(status.getTimeSeconds()));
+        }
+        else
+        {
+            timeText.setString("0");
+        }
     }
     else
     {
-        timeText.setString("0");
+        // Alternative scoring
+        timeText.setString(
+            lua.readVariable<string>(levelStatus.scoreOverride));
     }
 
     const auto getScaledCharacterSize = [&](const float size) {
--- a/src/SSVOpenHexagon/Core/HGGraphics.cpp
+++ b/src/SSVOpenHexagon/Core/HGGraphics.cpp
@@ -305,8 +305,7 @@ void HexagonGame::updateText()
     else
     {
         // Alternative scoring
-        timeText.setString(
-            lua.readVariable<string>(levelStatus.scoreOverride));
+        timeText.setString(lua.readVariable<string>(levelStatus.scoreOverride));
     }
 
     const auto getScaledCharacterSize = [&](const float size) {
@@ -339,19 +338,15 @@ void HexagonGame::updateText()
         const replay_file& rf = activeReplay->replayFile;
 
         os.str("");
+        os << formatTime(rf._played_score / 60.0);
 
-        if (levelStatus.scoreOverridden)
+        if(!levelStatus.scoreOverridden)
         {
-            os << formatTime(rf._played_score) << " BY "
-               << rf._player_name << '\n';
-        }
-        else
-        {
-            os << formatTime(rf._played_score / 60.0) << "s BY "
-               << rf._player_name << '\n';
+            os << "s";
         }
 
-        os << activeReplay->replayPackName << " - "
+        os << " BY " << rf._player_name << '\n'
+           << activeReplay->replayPackName << " - "
            << activeReplay->replayLevelName << " (" << rf._difficulty_mult
            << "x)";
 
--- a/src/SSVOpenHexagon/Core/HGGraphics.cpp
+++ b/src/SSVOpenHexagon/Core/HGGraphics.cpp
@@ -339,11 +339,12 @@ void HexagonGame::updateText()
         const replay_file& rf = activeReplay->replayFile;
 
         os.str("");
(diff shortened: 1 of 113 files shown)


src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -330,8 +340,16 @@ void HexagonGame::updateText()
...
+        if (levelStatus.scoreOverridden)
+        {
+            os << formatTime(rf._played_score) << " BY "
+               << rf._player_name << '\n';
+        }
+        else
+        {
+            os << formatTime(rf._played_score / 60.0) << "s BY "
+               << rf._player_name << '\n';
+        }
```
SuperV1234, line 352: You can extract the "BY ..." portion of the `os <<` streaming out of the if statement, as it's the same in both branches.

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -290,13 +290,23 @@ void HexagonGame::updateText()
...
+        {
+            timeText.setString("0");
+        }
     }
     else
     {
-        timeText.setString("0");
+        // Alternative scoring
+        timeText.setString(
+            lua.readVariable<string>(levelStatus.scoreOverride));
```
SuperV1234, line 309:
```suggestion
            lua.readVariable<std::string>(levelStatus.scoreOverride));
```

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -761,6 +761,39 @@ void HexagonGame::initLua_LevelControl()
...
+            try
+            {
+                levelStatus.scoreOverride = mVar;
+                levelStatus.scoreOverridden = true;
+                // Make sure we're not passing in a string
+                lua.executeCode(
+                    "if (type(" + mVar +
+                    ") ~= \"number\") then\n"
+                    "error(\"Score override must be a number value\")\n"
+                    "end");
```
SuperV1234, line 775: A [raw string literal](https://stackoverflow.com/questions/56710024/what-is-a-raw-string) here would help a lot with readability.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -761,6 +761,39 @@ void HexagonGame::initLua_LevelControl()
...
+            }
+            catch(const std::runtime_error& mError)
```
SuperV1234, line 777: When can this happen, exactly?

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -134,6 +134,11 @@ void HexagonGame::update(ssvu::FT mFT)
 
             player.updatePosition(*this, mFT);
 
+            status.accumulateFrametime(mFT);
+            if (levelStatus.scoreOverridden) 
+            {
+                status.updateCustomScore(lua.readVariable<float>(levelStatus.scoreOverride));
+            }
```
SuperV1234, line 141: Was there any reason why `accumulateFrametime` was down below before? Honest question, can't remember...

SuperV1234 (changes requested):
I like the idea, opens up some new possibilities. Some minor tweaks needed :)

src/SSVOpenHexagon/Core/HGUpdate.cpp
Morxemplum: I dunno, but I think I made this change in the code because it was supposed to fix a small error in the level time when it saves to the menu (considering it was slightly off from the time shown in the attempt).

src/SSVOpenHexagon/Core/HGScripting.cpp
Morxemplum: This should be happening every time the score is being overridden with a different Lua variable. We want to make sure that the variable we are using to override the score has to be some sort of number value. We do not want the score override to be another type like a table, string, boolean, etc. Mainly because those types will be very hard to sort and it will be very arbitrary in the leader boards.

src/SSVOpenHexagon/Core/HGGraphics.cpp
Morxemplum: Yeah, I can see how this implementation can be much cleaner. I can go ahead and fix this real quick.

src/SSVOpenHexagon/Core/HGScripting.cpp
Morxemplum: Copy that.

SuperV1234 (approved):
LGTM

PR #279, opened 2020-08-14
Levels that aren't selectable no longer show up in the menu

### LINKED ISSUE #236: [Must Have] - Levels still accessible with "selectable: false"
image

_thonk_

Steps to reproduce:
1. Go to any level property's JSON and change `"selectable"` to `false`.
2. Launch the game.

**Expected:** The level can no longer be selected.
**Actual:** The level is still selectable.

A/C:
1. Fix this bug.

Morxemplum:
Fixes #236.

### DIFF
--- a/src/SSVOpenHexagon/Global/Assets.cpp
+++ b/src/SSVOpenHexagon/Global/Assets.cpp
@@ -301,10 +301,12 @@ void HGAssets::loadLevelData(
     for(const auto& p : scanSingleByExt(mPath + "Levels/", ".json"))
     {
         LevelData levelData{ssvuj::getFromFile(p), mPath, mPackId};
-
-        const std::string assetId = mPackId + "_" + levelData.id;
-        levelDataIdsByPack[mPackId].emplace_back(assetId);
-        levelDatas.emplace(assetId, std::move(levelData));
+        if (levelData.selectable)
+        {
+            const std::string assetId = mPackId + "_" + levelData.id;
+            levelDataIdsByPack[mPackId].emplace_back(assetId);
+            levelDatas.emplace(assetId, std::move(levelData));
+        }
     }
 }
 void HGAssets::loadLocalProfiles()


SuperV1234 (approved):
LGTM

SuperV1234 (changes requested):
Hmm, on second thought, I'm not too sure about this. We probably still want to load the level in memory, but just not display it in the menu. Not sure if some Lua code could try to access `levelDatas` on an available but unselectable level, and fail/crash. 

What do you think?

Morxemplum:
Wouldn't the Lua code just not run at all if the level is unselectable? Or are you referring to when we start to re-implement the set level function that we would be able to set the level to a level that is unselectable? I think the assets of the level will still load, but the level file itself will get ignored. Right now I don't see a practical use for grabbing a level id in some code.

SuperV1234:
> Wouldn't the Lua code just not run at all if the level is unselectable? Or are you referring to when we start to re-implement the set level function that we would be able to set the level to a level that is unselectable? I think the assets of the level will still load, but the level file itself will get ignored. Right now I don't see a practical use for grabbing a level id in some code.

I was thinking of the latter. It also seems like "unselectable" right now is implemented as "don't load the level at all", which seems kinda misleading/weird, and might cause problems in the future. I'd rather have "unselectable" just hide the level from the menu selection screen, but not change the way the level is loaded at all.

Morxemplum:
What is weird is that this solution was done because this was how 1.92 handled unselectable levels, and even then if you played Baby Steps on 1.92 it was able to set the level to "More Baby Steps", an unselectable level. So unless if I see concrete proof that it truly isn't loading the id into the engine, I don't think this is the case.

Morxemplum:
Closing this pull request. With the new UI, I can find a completely different way to incorporate unselectable levels in a way that Vee likes. Since this pull request is so far behind on the actual code, there's no point in keeping it open.

PR #282, opened 2020-08-28
Boot OH into a level with command line argument

IvoryDuke:
Brief: add command line parameter -l and a level name to make OH load a level on boot. Mostly meant for level makers who might take advantage of not scrolling trough the menus to start the level they are working on.
https://youtu.be/9iyT4n7QacA

Details:
- add -l followed by the level name to load straight into the level (arguments used in the video are -l seconddimension);
- the level name is sent to the Menu class when MenuGame::Init() is called for the first time and here it is processed to see if there is a loaded level with matching name. In case no level is found an error message is printed to console and the load process is aborted;
- the level load will also abort if there is no local profile created, and the process will stop at the profile creation screen.

Notes:
- I store the level argument later compared to "overrideIds" in main.cpp because the string corrupts during the load process until it becomes an empty string;
- also removed a few redundant #includes.

### DIFF
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -241,7 +241,8 @@ public:
         Discord::discord_manager& mDiscordManager, HGAssets& mAssets,
         HexagonGame& mHexagonGame, ssvs::GameWindow& mGameWindow);
 
-    void init(bool mErrored);
+    void init(bool mErrored, const std::string& level = "");
+    void loadCommandLineLevel(const std::string& level);
 
     ssvs::GameState& getGame() noexcept
     {
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -241,8 +241,9 @@ public:
         Discord::discord_manager& mDiscordManager, HGAssets& mAssets,
         HexagonGame& mHexagonGame, ssvs::GameWindow& mGameWindow);
 
-    void init(bool mErrored, const std::string& level = "");
-    void loadCommandLineLevel(const std::string& level);
+    void init(bool mErrored);
+    void init(bool mErrored, const std::string& pack, const std::string& level);
+    void loadCommandLineLevel(const std::string& pack, const std::string& level);
 
     ssvs::GameState& getGame() noexcept
     {
--- a/src/SSVOpenHexagon/Core/HexagonGame.cpp
+++ b/src/SSVOpenHexagon/Core/HexagonGame.cpp
@@ -3,25 +3,12 @@
 // AFL License page: http://opensource.org/licenses/AFL-3.0
 
 #include "SSVOpenHexagon/Global/Assets.hpp"
-#include "SSVOpenHexagon/Global/Common.hpp"
 #include "SSVOpenHexagon/Core/HexagonGame.hpp"
 #include "SSVOpenHexagon/Core/MenuGame.hpp"
 #include "SSVOpenHexagon/Core/Joystick.hpp"
-#include "SSVOpenHexagon/Core/Steam.hpp"
-#include "SSVOpenHexagon/Core/Discord.hpp"
-#include "SSVOpenHexagon/Online/Online.hpp"
-#include "SSVOpenHexagon/Utils/Utils.hpp"
-#include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
-
-#include <SSVStart/Utils/Vector2.hpp>
-#include <SSVStart/SoundPlayer/SoundPlayer.hpp>
-
-#include <SSVUtils/Core/Common/Frametime.hpp>
-
 #include <cassert>
 
 using namespace hg::Utils;
-
 
 namespace hg
 {
--- a/src/SSVOpenHexagon/Core/HexagonGame.cpp
+++ b/src/SSVOpenHexagon/Core/HexagonGame.cpp
@@ -3,9 +3,20 @@
 // AFL License page: http://opensource.org/licenses/AFL-3.0
 
 #include "SSVOpenHexagon/Global/Assets.hpp"
+#include "SSVOpenHexagon/Global/Common.hpp"
 #include "SSVOpenHexagon/Core/HexagonGame.hpp"
 #include "SSVOpenHexagon/Core/MenuGame.hpp"
 #include "SSVOpenHexagon/Core/Joystick.hpp"
+#include "SSVOpenHexagon/Core/Steam.hpp"
+#include "SSVOpenHexagon/Core/Discord.hpp"
+#include "SSVOpenHexagon/Online/Online.hpp"
+#include "SSVOpenHexagon/Utils/Utils.hpp"
+#include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
+
+#include <SSVStart/Utils/Vector2.hpp>
+#include <SSVStart/SoundPlayer/SoundPlayer.hpp>
+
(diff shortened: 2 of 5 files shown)


src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -3,26 +3,13 @@
...
-#include "SSVOpenHexagon/Core/Steam.hpp"
-#include "SSVOpenHexagon/Core/Discord.hpp"
-#include "SSVOpenHexagon/Online/Online.hpp"
-#include "SSVOpenHexagon/Utils/Utils.hpp"
-#include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
-
-#include <SSVStart/Utils/Vector2.hpp>
-#include <SSVStart/SoundPlayer/SoundPlayer.hpp>
-
-#include <SSVUtils/Core/Common/Frametime.hpp>
```
SuperV1234, line 19: Please keep these. Relying on transitive includes is bad C++ practice.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -6,18 +6,6 @@
...
-#include "SSVOpenHexagon/Online/Online.hpp"
-#include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
-#include "SSVOpenHexagon/SSVUtilsJson/SSVUtilsJson.hpp"
-
-#include <SSVStart/Input/Input.hpp>
-#include <SSVStart/Utils/Vector2.hpp>
-
-#include <SSVMenuSystem/SSVMenuSystem.hpp>
-
-#include <SSVUtils/Core/Common/Frametime.hpp>
```
SuperV1234, line 9: Ditto

include/SSVOpenHexagon/Core/MenuGame.hpp
```diff
@@ -241,7 +241,8 @@ class MenuGame
         Discord::discord_manager& mDiscordManager, HGAssets& mAssets,
         HexagonGame& mHexagonGame, ssvs::GameWindow& mGameWindow);
 
-    void init(bool mErrored);
+    void init(bool mErrored, const std::string& level = "");
```
SuperV1234, line 244: Not a fan of this interface, because it relies on `level == ""` as a magic value to avoid loading a command-line level. We should have two separate functions here:

```cpp
void initWithLevel(bool mErrored, const std::string& level);
void initWithoutLevel(bool mErrored); 
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -70,15 +58,70 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
...
+void MenuGame::loadCommandLineLevel(const std::string& level)
+{
+    // retrieve the level index location
+    int levelIndex{-1}, i;
+    const auto& p(assets.getPackInfos());
+    std::vector<std::string> levelsIDs;
+    std::string packID, levelID;
+    std::vector<string>::iterator it;
+
+    for(i = 0; i < int(p.size()); ++i)
```
SuperV1234, line 78: Unless I am misunderstanding how this works, `level` is just the plain name of a level (e.g., "apeirogon"), and then you attempt to find any level named "apeirogon" in any of the installed packs by doing some string manipulation on the level id. 

This seems OK, but it does mean that it is not possible to properly handle cases where two separate packs have a level with the same name. Have you given any thought to this?

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -70,15 +58,70 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
...
 
-void MenuGame::init(bool error)
+void MenuGame::loadCommandLineLevel(const std::string& level)
+{
+    // retrieve the level index location
+    int levelIndex{-1}, i;
+    const auto& p(assets.getPackInfos());
+    std::vector<std::string> levelsIDs;
+    std::string packID, levelID;
+    std::vector<string>::iterator it;
```
SuperV1234, line 76:
```suggestion
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -70,15 +58,70 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
...
+
+    for(i = 0; i < int(p.size()); ++i)
+    {
+        packID = p.at(i).id;
+        levelID = packID + "_";
+        levelID += level;
+        levelsIDs = assets.getLevelIdsByPack(packID);
+        it = find(levelsIDs.begin(), levelsIDs.end(), levelID);
```
SuperV1234, line 84:
```suggestion
        auto it = find(levelsIDs.begin(), levelsIDs.end(), levelID);
```

src/SSVOpenHexagon/Core/main.cpp
```diff
@@ -8,12 +8,6 @@
 #include "SSVOpenHexagon/Online/OHServer.hpp"
 #include "SSVOpenHexagon/Core/HexagonGame.hpp"
 #include "SSVOpenHexagon/Core/MenuGame.hpp"
-#include "SSVOpenHexagon/Core/Steam.hpp"
-#include "SSVOpenHexagon/Core/Discord.hpp"
-#include "SSVOpenHexagon/Global/Assets.hpp"
-#include "SSVOpenHexagon/Global/Config.hpp"
-
-#include <SSVStart/GameSystem/GameWindow.hpp>
```
SuperV1234, line 16: Ditto, let's keep these

src/SSVOpenHexagon/Core/main.cpp
```diff
@@ -42,11 +36,15 @@ int main(int argc, char* argv[])
...
 
-    const auto overrideIds = [&] {
+    const auto overrideIds = [&]
+    {
         std::vector<std::string> result;
         for(int i{0}; i < argc; ++i)
         {
-            result.emplace_back(argv[i]);
+            if(!strcmp(argv[i], "-l")) // need this later
+                i += 2;
```
SuperV1234, line 45: Rather than looping again afterwards, you could store the found level here, maybe into a `std::optional<std::string>` so that you can also check later whether any `-l` flag was supplied or not. Happy to elaborate on Discord.

src/SSVOpenHexagon/Core/main.cpp
```diff
@@ -104,7 +102,22 @@ int main(int argc, char* argv[])
...
+          if(!strcmp(argv[i], "-l"))
+          {
+              levelResult = argv[++i];
+              break;
+          }
+      }
+      return levelResult;
+    }();
+
+    mg->init(false /* mError */, levelName);
```
SuperV1234, line 120: I would replace all of this with the `std::optional` solution above, and with the new suggested interface:

```cpp
std::optional<std::string> commandLineLevelArg;
// ...

if (commandLineLevelArg.has_value())
{
    mg->initWithLevel(false /* mError */, commandLineLevelArg.value());
}
else
{
    mg->initWithoutLevel(false /* mError */);
}
```

SuperV1234 (changes requested):
Valuable addition, code needs a bit of work :)

src/SSVOpenHexagon/Core/MenuGame.cpp
IvoryDuke: I originally required both a pack and level name to boot into a level.
But I found it to be rather tiring to require two values and thought mapmakers should already put some thought in not having a same named level.
On second thought I see not everybody can know all names so I will reimplement the first version

SuperV1234 (approved):
LGTM. Let's merge!

PR #283, opened 2020-08-29
Reload assets of level selected in the main menu

IvoryDuke:
Brief:
press F5 in the main menu to reload the assets of the currently selected level. To be used by pack developers so that they don't have to turn OH off and on to test small changes.
https://youtu.be/DPiZLw8cshQ

Details:
- press F5 in the main menu to reload assets of the level;
- function reload the level data and then music, style, and custom sounds files the level data .json is pointing to;
- error messages are printed in console if the file does not exist;
- files to be reloaded must be in the pack directory subfolders (ex. PackPath/Music/ or PackPath/Style/);
- new music and custom sound files are not reloaded if they have been already cached at boot.

Extra notes:
- removed some redundant includes;
- added soundId to LevelData class;
- slightly reorganized Assets.hpp and Assets.cpp

### DIFF
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -236,6 +236,8 @@ private:
         return getCurrentMenu() != nullptr;
     }
 
+    void reloadLevelAssets();
+
 public:
     MenuGame(Steam::steam_manager& mSteamManager,
         Discord::discord_manager& mDiscordManager, HGAssets& mAssets,
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -101,7 +101,8 @@ private:
     sf::Text txtVersion{"", imagine, 40}, txtProf{"", imagine, 21},
         txtLName{"", imagine, 65}, txtLDesc{"", imagine, 32},
         txtLAuth{"", imagine, 20}, txtLMus{"", imagine, 20},
-        txtFriends{"", imagine, 21}, txtPacks{"", imagine, 14};
+        txtFriends{"", imagine, 21}, txtPacks{"", imagine, 14},
+        txtReload{"", imagine, 32};
 
     void playLocally();
 
@@ -237,6 +238,7 @@ private:
     }
 
     void reloadLevelAssets();
+    float reloadMessageTimer{0.f};
 
 public:
     MenuGame(Steam::steam_manager& mSteamManager,
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -104,9 +104,12 @@ private:
         txtFriends{"", imagine, 21}, txtPacks{"", imagine, 14};
 
     // dialog box
+    void drawDialogBox();
+    Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogBackdrop;
+    Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogFrame;
+    std::vector<std::string> dialogText;
     sf::Text txtDialog{"", imagine, 0};
-    std::string dialogText;
-    float dialogHeight{0.f};
+    float dialogHeight{0.f}, dialogWidth{0.f}, frameOffset{0.f}, lineHeight{0.f};
     int noActions{0};
 
     void playLocally();
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -105,6 +105,7 @@ private:
 
     // dialog box
     void drawDialogBox();
+    void clearDialogBox();
     Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogBackdrop;
     Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogFrame;
     std::vector<std::string> dialogText;
@@ -251,6 +252,11 @@ public:
     MenuGame(Steam::steam_manager& mSteamManager,
         Discord::discord_manager& mDiscordManager, HGAssets& mAssets,
         HexagonGame& mHexagonGame, ssvs::GameWindow& mGameWindow);
+
+    ~MenuGame()
+    {
+        clearDialogBox();
+    }
 
     void init(bool mErrored);
 
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -45,7 +45,7 @@ public:
     void createDialogBox(std::string& output, const int charSize);
     void drawDialogBox();
     void clearDialogBox();
-    bool empty()
+    [[nodiscard]] bool empty() const noexcept
     {
         return dialogText.empty();
     }
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -102,7 +102,9 @@ private:
         txtLName{"", imagine, 65}, txtLDesc{"", imagine, 32},
         txtLAuth{"", imagine, 20}, txtLMus{"", imagine, 20},
         txtFriends{"", imagine, 21}, txtPacks{"", imagine, 14},
-        txtReload{"", imagine, 32};
+        txtDialog{"", imagine, 0};
+    bool dialogBox{false};
+    int noActions{0};
 
     void playLocally();
 
@@ -238,7 +240,6 @@ private:
     }
 
     void reloadLevelAssets();
-    float reloadMessageTimer{0.f};
 
 public:
     MenuGame(Steam::steam_manager& mSteamManager,
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -106,7 +106,6 @@ private:
     // dialog box
     void drawDialogBox();
     void clearDialogBox();
-    Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogBackdrop;
     Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogFrame;
     std::vector<std::string> dialogText;
     sf::Text txtDialog{"", imagine, 0};
@@ -252,11 +251,6 @@ public:
     MenuGame(Steam::steam_manager& mSteamManager,
         Discord::discord_manager& mDiscordManager, HGAssets& mAssets,
         HexagonGame& mHexagonGame, ssvs::GameWindow& mGameWindow);
-
-    ~MenuGame()
-    {
-        clearDialogBox();
-    }
 
     void init(bool mErrored);
 
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -26,7 +26,7 @@
 namespace hg
 {
 
-class OHDialogBox
+class HexagonDialogBox // can't be named DialogBox due to a pre-existing C++ class
 {
 private:
     HGAssets& assets;
@@ -41,7 +41,7 @@ private:
     float dialogHeight{0.f}, dialogWidth{0.f}, frameOffset{0.f}, lineHeight{0.f};
 
 public:
-    OHDialogBox(HGAssets& mAssets, ssvs::GameWindow& window, StyleData& styleData);
+    HexagonDialogBox(HGAssets& mAssets, ssvs::GameWindow& window, StyleData& styleData);
     void createDialogBox(std::string& output, const int charSize);
     void drawDialogBox();
     void clearDialogBox();
@@ -128,7 +128,7 @@ private:
         txtLAuth{"", imagine, 20}, txtLMus{"", imagine, 20},
         txtFriends{"", imagine, 21}, txtPacks{"", imagine, 14};
 
-    OHDialogBox dialogBox;
+    HexagonDialogBox dialogBox;
 
     void playLocally();
 
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -26,6 +26,31 @@
 namespace hg
 {
 
+class OHDialogBox
+{
+private:
+    HGAssets& assets;
+    ssvs::GameWindow& window;
+    StyleData& styleData;
+    sf::Font& imagine = assets.get<sf::Font>(
+        "forcedsquare.ttf"); // G++ bug (cannot initialize with curly braces)
+
+    Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogFrame;
+    std::vector<std::string> dialogText;
+    sf::Text txtDialog{"", imagine, 0};
+    float dialogHeight{0.f}, dialogWidth{0.f}, frameOffset{0.f}, lineHeight{0.f};
+
+public:
+    OHDialogBox(HGAssets& mAssets, ssvs::GameWindow& window, StyleData& styleData);
+    void createDialogBox(std::string& output, const int charSize);
+    void drawDialogBox();
+    void clearDialogBox();
+    bool empty()
+    {
+        return dialogText.empty();
+    }
+};
+
 enum class States
 {
     EpilepsyWarning,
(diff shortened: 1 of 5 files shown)


IvoryDuke:
Added on screen message with the reload outcome
Capture

SuperV1234:
This is a fantastic addition. Just came back to London, will review ASAP!

IvoryDuke:
Thank you!
A couple of notes I did not mention in the main body:
- I set up the message with a genericly named "dialogBox" cause the idea is that it can be used for any other messages that require a player input;
- the way actions are not executed when the dialog box is up would be more efficient with the "ignoreAllInputs" functions I added to my SSVStart pull request since it only requires the inputManager to be notified to ignore the inputs instead of placing a return statement in every single action. I did not inplement that in this here cause it would have meant to make the builds fail like in my "Menu Binds options" pull request. Although it would work the same: when dialog shows up set ignoreAllInputs(true) and set it back to false when the dialog is removed.

IvoryDuke:
Finalized the look of the reload message.
I think it looks prettier center aligned
Capture

IvoryDuke:
Added frame and backdrop to the dialog box for better looks and readability
https://youtu.be/r-0ACyM6FiA

include/SSVOpenHexagon/Core/MenuGame.hpp
```diff
@@ -103,6 +103,15 @@ class MenuGame
...
 
+    // dialog box
+    void drawDialogBox();
+    void clearDialogBox();
+    Utils::FastVertexVector<sf::PrimitiveType::Quads> dialogFrame;
+    std::vector<std::string> dialogText;
+    sf::Text txtDialog{"", imagine, 0};
+    float dialogHeight{0.f}, dialogWidth{0.f}, frameOffset{0.f}, lineHeight{0.f};
+    int noActions{0};
+
```
SuperV1234, line 114: How about putting all of this stuff into its own class? Seems quite self-contained, and it could be used for other dialog boxes

include/SSVOpenHexagon/Global/Assets.hpp
```diff
@@ -137,19 +137,24 @@ class HGAssets
...
     const MusicData& getMusicData(
         const std::string& mPackId, const std::string& mId);
     const StyleData& getStyleData(
         const std::string& mPackId, const std::string& mId);
 
+    std::string reloadLevelData(const std::string& mPackId, const std::string& mPath, const std::string& mId);
+    std::string reloadMusicData(const std::string& mPackId, const std::string& mPath, const std::string& mId);
+    std::string reloadStyleData(const std::string& mPackId, const std::string& mPath, const std::string& mId);
+    std::string reloadMusic(const std::string& mPackId, const std::string& mPath, const std::string& mId);
+    std::string reloadCustomSounds(const std::string& mPackId, const std::string& mPath, const std::string& mId);
```
SuperV1234, line 151: In general, functions whose return value should never be ignored should be marked as `[[nodiscard]]`. Despite the verbosity, this allows the compiler to produce a warning if someone forgets to check the return value. It's particularly useful here because the names do not suggest that a return value will be returned:

```suggestion
    [[nodiscard]] std::string reloadLevelData(const std::string& mPackId, const std::string& mPath, const std::string& mId);
    [[nodiscard]] std::string reloadMusicData(const std::string& mPackId, const std::string& mPath, const std::string& mId);
    [[nodiscard]] std::string reloadStyleData(const std::string& mPackId, const std::string& mPath, const std::string& mId);
    [[nodiscard]] std::string reloadMusic(const std::string& mPackId, const std::string& mPath, const std::string& mId);
    [[nodiscard]] std::string reloadCustomSounds(const std::string& mPackId, const std::string& mPath, const std::string& mId);
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -6,18 +6,6 @@
...
-#include "SSVOpenHexagon/Online/Online.hpp"
-#include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
-#include "SSVOpenHexagon/SSVUtilsJson/SSVUtilsJson.hpp"
-
-#include <SSVStart/Input/Input.hpp>
-#include <SSVStart/Utils/Vector2.hpp>
-
-#include <SSVMenuSystem/SSVMenuSystem.hpp>
-
-#include <SSVUtils/Core/Common/Frametime.hpp>
```
SuperV1234, line 20: We should keep these. It's always better to be explicit about includes. Relying on transitive includes makes refactoring way harder, because now modifying `HexagonGame.hpp` might break `MenuGame.cpp` :)

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -68,6 +56,24 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
...
+        {
+			if(!noActions) return;
+            if(!(--noActions)) clearDialogBox();
+        };
+    game.onEvent(Event::EventType::JoystickButtonReleased) +=
+        [this](const Event& mEvent)
+        {
+          if(!noActions) return;
+          if(!(--noActions)) clearDialogBox();
+        };
```
SuperV1234, line 76:
```suggestion
    const auto clearBoxIfActionNeeded = [this](const Event& mEvent)
    {
        if(!noActions) return;
        if(!(--noActions)) clearDialogBox();
    };

    game.onEvent(Event::EventType::KeyReleased) += clearBoxIfActionNeeded;
    game.onEvent(Event::EventType::MouseButtonReleased) += clearBoxIfActionNeeded; 
    game.onEvent(Event::EventType::JoystickButtonReleased) += clearBoxIfActionNeeded;
```

I think the above should work and help reduce repetition. Also, some comments would be nice in the body of the lambda expression, it is not immediately clear what `!(--noActions)` means (not syntactically, but in terms of what it means for the player)

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -402,6 +421,8 @@ void MenuGame::upAction()
 }
 void MenuGame::downAction()
 {
+    if(noActions) return;
```
SuperV1234, line 424: Again, this is not immediately clear. Maybe rename `noActions` to something more specific? Maybe something like `suppressActionsDueToDialogBox`?

SuperV1234: Also, sorry... but I like my braces :)

```suggestion
    if(noActions) { return; }
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -611,19 +645,25 @@ void MenuGame::initInput()
...
         Config::getTriggerExit(),
-        [this](ssvu::FT mFT) {
-            if(state != States::MOpts)
-            {
-                exitTimer += mFT;
-            }
+        [this](ssvu::FT mFT)
+        {
+            if(noActions) return;
+            if(state != States::MOpts) exitTimer += mFT;
```
SuperV1234, line 651: Ditto, please do not alter existing code style. Not a big deal, but it's just going to be reverted...

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -638,6 +678,56 @@ void MenuGame::initInput()
...
+    game.addInput(
+        {{k::F5}},
+        [this](ssvu::FT /*unused*/) { reloadLevelAssets(); },t::Once);
+}
+
+void MenuGame::reloadLevelAssets()
+{
+    if(state != States::SMain || !dialogText.empty() || !Config::getDebug()) return;
+    
+    noActions = 2;
```
SuperV1234, line 690: This line needs to be more self-explanatory or needs a comment. Why `2`? What does this mean in practice?

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -638,6 +678,56 @@ void MenuGame::initInput()
...
+    assets.playSound("beep.ogg");
+
+    string reloadOutput = assets.reloadLevelData(levelData->packId, levelData->packPath, levelData->id);
+    if(reloadOutput != "invalid level folder path\n" && reloadOutput != "no matching level data file found\n")
```
SuperV1234, line 694: Hmm, it seems that here you are relying on the string output of `reloadLevelData` to decide what to do. This is brittle, because someone might change the return value of `reloadOutput` assuming that it is just used as a human-readable message, while you're actually using it programmatically in here. 

Generally, to observe multiple possibilities returned from a function, `enum class` is used in C++. If you need to carry some additional data per-possibility, you can use `std::variant` instead. Happy to give you some examples on Discord, if I am not misunderstanding the purpose of this code.

SuperV1234: I think you want both a `reloadOutput` string message for the human, and some sort of `reloadOutcome` enum/variant for the program here.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -638,6 +678,56 @@ void MenuGame::initInput()
...
+
+        if(levelData->musicId != "nullMusicId")
+            reloadOutput += assets.reloadMusic(levelData->packId, levelData->packPath, levelData->musicId);
+
+        if(levelData->soundId != "nullSoundId")
+            reloadOutput += assets.reloadCustomSounds(levelData->packId, levelData->packPath, levelData->soundId);
+    }
+
+    reloadOutput += "\n";
+    reloadOutput += "press any key to close this message\n";
```
SuperV1234, line 709:
```suggestion
    reloadOutput += "\npress any key to close this message\n";
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -638,6 +678,56 @@ void MenuGame::initInput()
...
+    Utils::uppercasify(reloadOutput);
+
+    txtDialog.setCharacterSize(26);
+    txtDialog.setString(reloadOutput);
+    dialogWidth = getGlobalWidth(txtDialog);
+    dialogHeight = getGlobalHeight(txtDialog);
+    txtDialog.setString("A");
```
SuperV1234, line 716: What's the point of this? Comment would be nice.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -638,6 +678,56 @@ void MenuGame::initInput()
...
+    lineHeight = getGlobalHeight(txtDialog);
+    frameOffset = 10.f;
+
+    string temp;
```
SuperV1234, line 720:
```suggestion
    std::string temp;
```

Let's try to explicitly qualify classes we use, it helps avoid name collisions and make code easier to move around.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -638,6 +678,56 @@ void MenuGame::initInput()
...
+}
+
+void MenuGame::reloadLevelAssets()
+{
+    if(state != States::SMain || !dialogText.empty() || !Config::getDebug()) return;
+    
+    noActions = 2;
+    assets.playSound("beep.ogg");
+
+    string reloadOutput = assets.reloadLevelData(levelData->packId, levelData->packPath, levelData->id);
```
SuperV1234, line 693:
```suggestion
    std::string reloadOutput = assets.reloadLevelData(levelData->packId, levelData->packPath, levelData->id);
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -638,6 +678,56 @@ void MenuGame::initInput()
...
+    txtDialog.setCharacterSize(26);
+    txtDialog.setString(reloadOutput);
+    dialogWidth = getGlobalWidth(txtDialog);
+    dialogHeight = getGlobalHeight(txtDialog);
+    txtDialog.setString("A");
+    lineHeight = getGlobalHeight(txtDialog);
+    frameOffset = 10.f;
+
+    string temp;
+    for(auto& c : reloadOutput)
```
SuperV1234, line 721:
```suggestion
    for(char c : reloadOutput)
```

Using `auto` and a reference here are both unnecessary and just obfuscate the fact that `c` is a `char`.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -1190,6 +1280,60 @@ void MenuGame::draw()
     {
         window.setMouseCursorVisible(hg::Config::getMouseVisible());
     }
+
+    if(!dialogText.empty())
+        drawDialogBox();
+}
+
+void MenuGame::drawDialogBox()
```
SuperV1234, line 1288: I strongly feel like all of the dialog box related data members and member functions should be encapsulated in a `DialogBox` class.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -1190,6 +1280,60 @@ void MenuGame::draw()
...
+    p1 = {leftBorder - frameOffset, halfH - dialogHeight - frameOffset + heightDif};  // top left
+    p2 = {rightBorder + frameOffset, halfH - dialogHeight - frameOffset + heightDif}; // top right
+    p3 = {rightBorder + frameOffset, halfH + frameOffset - heightDifBottom};          // bottom right
+    p4 = {leftBorder - frameOffset, halfH + frameOffset - heightDifBottom};           // bottom left
+    dialogFrame.reserve_more(4);
+    dialogFrame.batch_unsafe_emplace_back(styleData.getColor(0), p1, p2, p3, p4);
+    render(dialogFrame);
+
+    // text
+    float heightOffsset = 0.f;
```
SuperV1234, line 1323: Typo: "offset"

SuperV1234 (changes requested):
A very good start, and much wanted feature! Thanks. Needs a bit of extra work before it's ready to be merged.

src/SSVOpenHexagon/Core/MenuGame.cpp
IvoryDuke: I see, i went with that formatting cause it is used in uppercasify
`inline void uppercasify(std::string& s)
{
    for(auto& c : s)
    {
        c = std::toupper(c);
    }
}`

src/SSVOpenHexagon/Core/MenuGame.cpp
IvoryDuke: I totally agree, I will take care of it

src/SSVOpenHexagon/Core/MenuGame.cpp
InfoTeddy: Formatting code blocks isn't that hard. You type three backticks on one line, then the language name, and then at the end of it you type three backticks again. Like this:
```c++
inline void uppercasify(std::string& s)
{
    for(auto& c : s)
    {
        c = std::toupper(c);
    }
}
```

IvoryDuke: Got it

include/SSVOpenHexagon/Core/MenuGame.hpp
```diff
@@ -26,6 +26,31 @@
 namespace hg
 {
 
+class HexagonDialogBox // can't be named DialogBox due to a pre-existing C++ class
```
SuperV1234, line 29: This is surprising. Where is this `DialogBox` coming from?

include/SSVOpenHexagon/Core/MenuGame.hpp
```diff
@@ -26,6 +26,31 @@
...
+
+public:
+    HexagonDialogBox(HGAssets& mAssets, ssvs::GameWindow& window, StyleData& styleData);
+    void createDialogBox(std::string& output, const int charSize);
+    void drawDialogBox();
+    void clearDialogBox();
+    bool empty()
+    {
+        return dialogText.empty();
+    }
```
SuperV1234, line 51:
```suggestion
    [[nodiscard]] bool empty() const noexcept
    {
        return dialogText.empty();
    }
```

src/SSVOpenHexagon/Global/Assets.cpp
```diff
@@ -3,18 +3,9 @@
...
-#include "SSVOpenHexagon/Online/Definitions.hpp"
-#include "SSVOpenHexagon/Online/Online.hpp"
 #include "SSVOpenHexagon/Utils/Utils.hpp"
-#include "SSVOpenHexagon/Data/MusicData.hpp"
-#include "SSVOpenHexagon/SSVUtilsJson/SSVUtilsJson.hpp"
 #include "SSVOpenHexagon/Core/Steam.hpp"
 
-#include <SSVStart/SoundPlayer/SoundPlayer.hpp>
-
-#include <SSVUtils/Core/FileSystem/FileSystem.hpp>
```
SuperV1234, line 16: Please restore these

SuperV1234 (changes requested):
Almost ready. I'll clean up the last few things

include/SSVOpenHexagon/Core/MenuGame.hpp
IvoryDuke: I was half wrong in my statement, it is not a class but a MinGW define
```
#define DialogBox __MINGW_NAME_AW(DialogBox)
```
This is what pops up in my code editor

src/SSVOpenHexagon/Global/Assets.cpp
IvoryDuke: Done, I will keep in mind to not remove headers in the future

SuperV1234 (approved):
LGTM. Will do some refactoring myself. Thanks!

PR #284, opened 2020-08-30
Lua function to edit the player movement speed

IvoryDuke:
Brief:
use l_setPlayerSpeedMult() to apply a multiplier to the player speed. If not used, by default the multiplier is 1.f. l_getPlayerSpeedMult() returns the value.

Details:
- when player is created a reference to the LevelStatus value playerSpeedMult is passed, so that it can be checked upon during the level;
- the multiplier is applied onto the player speed in CPlayer::updateInput().

Extra notes:
removed a few redundant includes.

### DIFF
--- a/include/SSVOpenHexagon/Components/CPlayer.hpp
+++ b/include/SSVOpenHexagon/Components/CPlayer.hpp
@@ -31,6 +31,7 @@ private:
     float size;
     float speed;
     float focusSpeed;
+    float* playerSpeedMult;
 
     bool dead;
     bool justSwapped;
@@ -43,7 +44,7 @@ private:
     void drawDeathEffect(HexagonGame& mHexagonGame);
 
 public:
-    CPlayer(const sf::Vector2f& mStartPos, const float swapCooldown) noexcept;
+    CPlayer(const sf::Vector2f& mPos, const float swapCooldown, float* mPlayerSpeedMulti) noexcept;
 
     [[gnu::always_inline, nodiscard]] const sf::Vector2f&
     getPosition() const noexcept
--- a/include/SSVOpenHexagon/Components/CPlayer.hpp
+++ b/include/SSVOpenHexagon/Components/CPlayer.hpp
@@ -31,7 +31,6 @@ private:
     float size;
     float speed;
     float focusSpeed;
-    float* playerSpeedMult;
 
     bool dead;
     bool justSwapped;
@@ -44,7 +43,7 @@ private:
     void drawDeathEffect(HexagonGame& mHexagonGame);
 
 public:
-    CPlayer(const sf::Vector2f& mPos, const float swapCooldown, float* mPlayerSpeedMulti) noexcept;
+    CPlayer(const sf::Vector2f& mPos, const float swapCooldown) noexcept;
 
     [[gnu::always_inline, nodiscard]] const sf::Vector2f&
     getPosition() const noexcept
--- a/include/SSVOpenHexagon/Components/CPlayer.hpp
+++ b/include/SSVOpenHexagon/Components/CPlayer.hpp
@@ -43,7 +43,7 @@ private:
     void drawDeathEffect(HexagonGame& mHexagonGame);
 
 public:
-    CPlayer(const sf::Vector2f& mPos, const float swapCooldown) noexcept;
+    CPlayer(const sf::Vector2f& mStartPos, const float swapCooldown) noexcept;
 
     [[gnu::always_inline, nodiscard]] const sf::Vector2f&
     getPosition() const noexcept
--- a/src/SSVOpenHexagon/Components/CPlayer.cpp
+++ b/src/SSVOpenHexagon/Components/CPlayer.cpp
@@ -10,11 +10,11 @@ namespace hg
 
 inline constexpr float baseThickness{5.f};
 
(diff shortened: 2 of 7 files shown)


include/SSVOpenHexagon/Components/CPlayer.hpp
```diff
@@ -31,6 +31,7 @@ class CPlayer
     float size;
     float speed;
     float focusSpeed;
+    float* playerSpeedMult;
```
SuperV1234, line 34: Any reason why you designed this PR the way it is? It seems unusual that the player object itself holds a pointer to its speed multiplayer (see below...)

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -278,7 +264,7 @@ void CPlayer::updateInput(HexagonGame& mHexagonGame, ssvu::FT mFT)
     const int movement{mHexagonGame.getInputMovement()};
 
     const float currentSpeed =
-        mHexagonGame.getInputFocused() ? focusSpeed : speed;
+        *playerSpeedMult * (mHexagonGame.getInputFocused() ? focusSpeed : speed);
```
SuperV1234, line 267: I think this would make more sense if it was reading the player speed multiplayer directly from `mHexagonGame`, e.g.:

```cpp
const float currentSpeed =
    mHexagonGame.getPlayerSpeedMult() * (mHexagonGame.getInputFocused() ? focusSpeed : speed);
```

That way we avoid the need of passing an extra pointer around to the player.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -416,6 +416,11 @@ void HexagonGame::initLua_LevelControl()
         "all walls immediately, and changes apply as soon as the next wall "
         "is created.");
 
+    lsVar("PlayerSpeedMult", &LevelStatus::playerSpeedMult,
+          "Gets the speed multiplier of the player",
+
+          "Sets the speed multiplier of the player");
+
```
SuperV1234, line 423:
```suggestion
    lsVar("PlayerSpeedMult", &LevelStatus::playerSpeedMult,
          "Gets the speed multiplier of the player.",
          "Sets the speed multiplier of the player.");
          
```

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -288,10 +287,12 @@ void HexagonGame::newGame(const std::string& mPackId, const std::string& mId,
 
     lua = Lua::LuaContext{};
     initLua();
-
     runLuaFile(levelData->luaScriptPath);
     runLuaFunction<void>("onInit");
     runLuaFunction<void>("onLoad");
+
+    player = CPlayer{ssvs::zeroVec2f, getSwapCooldown(), &levelStatus.playerSpeedMult};
+
```
SuperV1234, line 295: With my proposed design, there's also no need to move the initialization of player after `onLoad`

SuperV1234 (changes requested):

IvoryDuke:
I totally agree, I will redesign it like that

include/SSVOpenHexagon/Components/CPlayer.hpp
```diff
@@ -43,7 +43,7 @@ class CPlayer
     void drawDeathEffect(HexagonGame& mHexagonGame);
 
 public:
-    CPlayer(const sf::Vector2f& mStartPos, const float swapCooldown) noexcept;
+    CPlayer(const sf::Vector2f& mPos, const float swapCooldown) noexcept;
```
SuperV1234, line 46:
```suggestion
    CPlayer(const sf::Vector2f& mStartPos, const float swapCooldown) noexcept;
```

No need for this change

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -3,21 +3,7 @@
...
-#include "SSVOpenHexagon/Global/Config.hpp"
-
-#include <SSVStart/Utils/SFML.hpp>
-#include <SSVStart/Utils/Vector2.hpp>
-
-#include <SSVUtils/Core/Common/Frametime.hpp>
-#include <SSVUtils/Core/Utils/Math.hpp>
-
-#include <SFML/System/Vector2.hpp>
-#include <SFML/Graphics/Color.hpp>
```
SuperV1234, line 20: Please restore these

SuperV1234 (changes requested):
Almost ready, two minor things

include/SSVOpenHexagon/Components/CPlayer.hpp
IvoryDuke: Oh my sorry! I don't remember why I did that but will totally revert

src/SSVOpenHexagon/Components/CPlayer.cpp
IvoryDuke: Done

include/SSVOpenHexagon/Components/CPlayer.hpp
IvoryDuke: Actually I now do remember why I did it.
I made the name consistent with the .cpp naming
```
CPlayer::CPlayer(const sf::Vector2f& mPos, const float swapCooldown) : [...]
```
I made both mStartPos hoping that is fine

SuperV1234 (approved):
LGTM, merging

PR #285, opened 2020-09-01
Fix pack list in main menu shifting horizontally when longest named pack is selected

IvoryDuke:
As the title states.

The issue is caused by the fact that the position of the text to be rendered is calculated depending on the width of the longest line.
Although this length has a different value when the longest line is selected because extra the selected pack indicator ">>> " takes more horizontal space than "    ".

To solve the problem, while the string of the pack list is being created the longest line is found and stored, and the horizontal offset of the list is calculates depending on its length.

### DIFF
--- a/src/SSVOpenHexagon/Core/MenuGame.cpp
+++ b/src/SSVOpenHexagon/Core/MenuGame.cpp
@@ -1320,24 +1320,22 @@ void MenuGame::drawLevelSelection()
         "(" + toStr(currentIndex + 1) + "/" + toStr(levelDataIds.size()) + ")",
         txtLMus, {20.f, getGlobalTop(lname) - 30.f});
 
-    string packNames{"Installed packs:\n"}, longestPackName;
+    string packNames{"Installed packs:\n"}, curPack, longestPackName;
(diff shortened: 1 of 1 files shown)


src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -1320,30 +1320,34 @@ void MenuGame::drawLevelSelection()
         "(" + toStr(currentIndex + 1) + "/" + toStr(levelDataIds.size()) + ")",
         txtLMus, {20.f, getGlobalTop(lname) - 30.f});
 
-    std::string packNames{"Installed packs:\n"};
+    string packNames{"Installed packs:\n"}, curPack, longestPackName;
```
SuperV1234, line 1323: In general, please do not change the coding style (e.g. getting rid of `std::` or of curly braces) -- I will revert these changes prior to merging anyway.

SuperV1234 (approved):
LGTM! Thanks

IvoryDuke:
I understand, I will keep it in mind for any future submissions

PR #287, opened 2020-09-11
Fixed compilation error, small optimization

IvoryDuke:
There was a double declaration of int noActions.

Removed
```
if(noActions)
{
    return;
}
```
from the single action() functions and took advantage of the newly added ignoreAllInputs(bool) function instead (already used for the menu bind inputting and applied it to DialogBox as well).


IvoryDuke:
The build still fails because the SSVStart must be updated

IvoryDuke:
Reworked in another pull request

PR #288, opened 2020-09-11
Use ignoreAllInputs instead of multiple if(noActions) { return; }

IvoryDuke:
Small optimization to use the recently added ignoreAllInputs GameState function

### DIFF
--- a/src/SSVOpenHexagon/Core/MenuGame.cpp
+++ b/src/SSVOpenHexagon/Core/MenuGame.cpp
@@ -81,6 +81,19 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
             return;
         }
 
+        // Scenario one: actions are blocked cause a dialog box is opened
+        if(!dialogBox.empty())
+        {
+            if(!(--noActions))
+            {
+                dialogBox.clearDialogBox();
+                game.ignoreAllInputs(false);
+                hg::Joystick::ignoreAllPresses(false);
+            }
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -81,6 +81,19 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
...
 
+        // Scenario one: actions are blocked cause a dialog box is opened
+        if(!dialogBox.empty())
+        {
+            if(!(--noActions))
+            {
+                dialogBox.clearDialogBox();
+                game.ignoreAllInputs(false);
+                hg::Joystick::ignoreAllPresses(false);
+            }
```
SuperV1234, line 92: Can you please extract this to a private function or lambda? It's repeated a few times

SuperV1234 (changes requested):

SuperV1234:
Also, @IvoryDuke, please rebase on top of the current `master` -- I did apply some changes before I saw this, unfortunately...

src/SSVOpenHexagon/Core/MenuGame.cpp
IvoryDuke: Sure will do

IvoryDuke:
Not a problem will take care of it.

IvoryDuke:
@SuperV1234 All done

SuperV1234 (approved):

PR #289, opened 2020-09-12
Fix onUnload's priority

Morxemplum:
This is a minor shift in the code so that way onUnload is now called after an initial runthrough of the entire Lua file. This is done to help ensure that onUnload is used for the purpose of changing level features after a first playthrough of the level, like intended.

### DIFF
--- a/include/SSVOpenHexagon/Core/Replay.hpp
+++ b/include/SSVOpenHexagon/Core/Replay.hpp
@@ -113,7 +113,7 @@ struct replay_file
     replay_data _data;        // Input data.
     std::string _pack_id;     // Id of the selected pack.
     std::string _level_id;    // Id of the played level.
-    bool _first_play;		  // If this was achieved on first level play
+    bool _first_play;		// If this was achieved on first level play.
     float _difficulty_mult;   // Played difficulty multiplier.
     double _played_score; // Played score (This can be an overridden score or
                           // frametime, excluding pauses).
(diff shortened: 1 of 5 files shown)


Morxemplum:
The newest commit I just made should close the issue that I just raised.

include/SSVOpenHexagon/Core/Replay.hpp
```diff
@@ -113,6 +113,7 @@ struct replay_file
     replay_data _data;        // Input data.
     std::string _pack_id;     // Id of the selected pack.
     std::string _level_id;    // Id of the played level.
+    bool _first_play;		  // If this was achieved on first level play
```
SuperV1234, line 116:
```suggestion
    bool _first_play;		// If this was achieved on first level play.
```

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -293,6 +295,7 @@ void HexagonGame::newGame(const std::string& mPackId, const std::string& mId,
         activeReplay->replayLevelName = Utils::toUppercase(levelData->name);
 
         rng = random_number_generator{activeReplay->replayFile._seed};
+		firstPlay = activeReplay->replayFile._first_play;
```
SuperV1234, line 298:
```suggestion
        firstPlay = activeReplay->replayFile._first_play;
```

SuperV1234 (approved):
Good catch with the bug. LGTM

PR #291, opened 2020-09-13
Keyboard bind menu item bugfix

IvoryDuke:
When a new key was assigned to a bind slot already taken by another key, the latter one was not wiped beforehand, causing load/read/save config issues.
This is what happened to the config.json file
```
"t_rotate_ccw" : 
[
	[ "kD", "kP" ], // causes issues
	[ "kC" ]
],
```


SuperV1234 (approved):
LGTM

PR #294, opened 2020-09-19
Bind controls beta testers suggestions

IvoryDuke:
- Increased keys per bind cap from 2 to 4;
- allowed assigning a same key to multiple binds;
- fixed an issue that caused OH to crash when the config was saved closing the game (fix in UtilJson.hpp, solution looks cheap and something else might fit better);
- added "t_select" to config.json, and added such bind customization in the menu;
- removed hardcoded keys bind prohibition;
- removed a slew of now useless Config functions;
- limited hardcoded key actions to Enter, Escape, Up, Down, Right, and Left.

### DIFF
--- a/include/SSVOpenHexagon/Global/UtilsJson.hpp
+++ b/include/SSVOpenHexagon/Global/UtilsJson.hpp
@@ -66,7 +66,8 @@ struct Converter<ssvs::KKey>
     {
         if(mValue == T::Unknown)
         {
-            arch(mObj, "");
+            std::string empty;
+            arch(mObj, empty);
             return;
         }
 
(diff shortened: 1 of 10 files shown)


include/SSVOpenHexagon/Global/UtilsJson.hpp
```diff
@@ -66,7 +66,8 @@ struct Converter<ssvs::KKey>
     {
         if(mValue == T::Unknown)
         {
-            arch(mObj, "");
+            std::string empty;
+            arch(mObj, empty);
```
SuperV1234, line 70:
```suggestion
            std::string empty;
            arch(mObj, empty);  // TODO: using `""` seems to be bugged
```

SuperV1234 (approved):
LGTM

PR #295, opened 2020-09-20
Change game version to follow Major.Minor.Micro format

### LINKED ISSUE #275: Represent Game Version in "Major.Minor.Micro" instead of a float
I originally closed my pull request #271 because I was thinking that I should probably redefine how the version is done and displayed before I start working on this again.

I think this is one of the more crucial things we need to do for the game overall. The current way of how versions are done is through a literal float value, which may be simple, but it can lead to many issues down the line when it comes to formatting or having an unexpected update come so soon. It'll allow us to make more smaller updates before we can officially declare bigger updates. Doing this would actually benefit us a lot, and we could consider the next release of Open Hexagon to be "2.0.3" instead of "2.03".

This may sound easy, but when I actually tried to accomplish this, I didn't realize how deeply ingrained the version is with the source code. It is heavily used for score checking (mostly just online functionalities) and it's because of this that makes this so much harder to pull off. I've made a structure for this new Version type so far, **but I am having issues with operator overloading**, which is something that is needed because this will be needed to compare older versions of the game to newer versions and accomplish the score checking code. If it weren't for the operator overloading, this would've been done and a pull request would've already been made. But I think maybe if I learn a bit more of operator overloading and how it works, then I'll probably be able to get this done.

This is definitely something we shouldn't discard, because I think it will majorly benefit us.

Morxemplum:
Closes #275.

Instead of the next Open Hexagon being Open Hexagon 2.03, it will now be referred to as Open Hexagon 2.0.3. I wanted to do this PR much earlier, but the Online code made that very difficult to work with. There's an entire struct and constant that is dedicated to the current version of the game, and there were a few other behind the scenes things that had to be changed to make this work properly.

In addition, I made a Lua function to get the current version of the game in a string called u_getVersion(). The utils lua file now has a new utility function called "atRequiredVersion", which is a useful utility function for pack developers who are developing packs on beta branches, to ensure that if they decide to release their pack early, people who aren't on the beta branch will know that they cannot run it.


include/SSVOpenHexagon/Data/ProfileData.hpp
```diff
@@ -11,21 +11,21 @@ namespace hg
...
 private:
-    float version;
+    std::string version;
     std::string name;
     ssvuj::Obj scores;
     std::vector<std::string> trackedNames;
 
 public:
-    ProfileData(float mVersion, const std::string& mName,
+    ProfileData(std::string mVersion, const std::string& mName,
```
SuperV1234, line 20:
```suggestion
    ProfileData(const std::string& mVersion, const std::string& mName,
```

include/SSVOpenHexagon/Data/ProfileData.hpp
```diff
@@ -11,21 +11,21 @@ namespace hg
...
         const ssvuj::Obj& mScores,
         const std::vector<std::string>& mTrackedNames)
         : version{mVersion}, name{mName}, scores{mScores}, trackedNames{
                                                                mTrackedNames}
     {
     }
 
-    [[nodiscard]] float getVersion() const
+    [[nodiscard]] std::string getVersion() const
```
SuperV1234, line 28:
```suggestion
    [[nodiscard]] const std::string& getVersion() const
```

include/SSVOpenHexagon/Global/Config.hpp
```diff
@@ -19,6 +20,39 @@ class Trigger;
...
+    int minor;
+    int micro;
+
+    Version& operator=(const Version& o)
+    {
+        major = o.major;
+        minor = o.minor;
+        micro = o.micro;
+        return *this;
+    }
```
SuperV1234, line 35:
```suggestion
```

This should be automatically generated by the compiler.

include/SSVOpenHexagon/Global/Config.hpp
```diff
@@ -19,6 +20,39 @@ class Trigger;
...
+
+    auto operator<(const Version& rhs) const
```
SuperV1234, line 37:
```suggestion
    [[nodiscard]] bool operator<(const Version& rhs) const
```

include/SSVOpenHexagon/Global/Config.hpp
```diff
@@ -19,6 +20,39 @@ class Trigger;
...
+    {
+        return std::tie(major, minor, micro) <
+               std::tie(rhs.major, rhs.minor, rhs.micro);
+    }
+
+    auto operator>(const Version& rhs) const
```
SuperV1234, line 43:
```suggestion
    [[nodiscard]] bool operator>(const Version& rhs) const
```

include/SSVOpenHexagon/Global/Config.hpp
```diff
@@ -19,6 +20,39 @@ class Trigger;
...
+    {
+        return std::tie(major, minor, micro) >
+               std::tie(rhs.major, rhs.minor, rhs.micro);
+    }
+
+    bool operator==(Version other) const
```
SuperV1234, line 49:
```suggestion
    [[nodiscard]] bool operator==(Version other) const
```

include/SSVOpenHexagon/Global/Config.hpp
```diff
@@ -19,6 +20,39 @@ class Trigger;
...
+    {
+        return (major == other.major) && (minor == other.minor) &&
+               (micro == other.micro);
+    }
+};
+const Version GAME_VERSION = Version{2, 0, 3};
```
SuperV1234, line 55:
```suggestion
constexpr Version GAME_VERSION{2, 0, 3};
```

src/SSVOpenHexagon/Global/Config.cpp
```diff
@@ -646,14 +646,17 @@ void setSaveLocalBestReplayToFile(bool mX)
     return fullscreen();
 }
 
-[[nodiscard, gnu::const]] float getVersion()
+[[nodiscard, gnu::const]] Version getVersion()
```
SuperV1234, line 649:
```suggestion
[[nodiscard, gnu::const]] constexpr Version getVersion()
```

src/SSVOpenHexagon/Global/Config.cpp
```diff
@@ -646,14 +646,17 @@ void setSaveLocalBestReplayToFile(bool mX)
...
+    return GAME_VERSION;
 }
 
 [[nodiscard, gnu::const]] const char* getVersionString()
 {
-    return "2.03";
+    return (std::to_string(GAME_VERSION.major) + "." +
+            std::to_string(GAME_VERSION.minor) + "." +
+            std::to_string(GAME_VERSION.micro))
+        .c_str();
```
SuperV1234, line 659: This is **undefined behavior**, as you're creating a temporary string and then returning `.c_str()`, which will be a pointer to some memory location which will immediately become invalid after returning from the function.

I would probably change this function to:

```cpp
[[nodiscard]] const std::string& getVersionString()
{
    static std::string result{
        std::to_string(GAME_VERSION.major) + "." +
        std::to_string(GAME_VERSION.minor) + "." +
        std::to_string(GAME_VERSION.micro)
    };

    return result;
}
```

The `static` guarantees that the string will be constructed once in a thread-safe way, and that will live for the entire duration of the program.

SuperV1234: You can also then remove the `std::string{...}` constructions from other places like in `u_getVersion`.

SuperV1234 (changes requested):
There's a bug to fix here, but generally looks good.

include/SSVOpenHexagon/Global/Config.hpp
Morxemplum: Interesting. Did not know this.

include/SSVOpenHexagon/Global/Config.hpp
Morxemplum: constexpr is new to me. I don't know what this is but I will take your word for it.

src/SSVOpenHexagon/Global/Config.cpp
Morxemplum: I was just confused because the type was changed from std::string to a const char*, which behaves differently from a string. I was trying to use the .c_str() to try and get a character pointer from a string, but I guess this works better than const char*

include/SSVOpenHexagon/Global/Config.hpp
Morxemplum: There's now a compiler error when this is applied. Again I have no knowledge on constexpr so I don't know what is going on with this error.

src/SSVOpenHexagon/Global/Config.cpp
Morxemplum: Alright rewrote the implementation using this suggestion, and made necessary changes in other parts of the code.

include/SSVOpenHexagon/Data/ProfileData.hpp
```diff
@@ -11,21 +11,21 @@ namespace hg
 class ProfileData
 {
 private:
-    float version;
+    std::string version;
```
SuperV1234, line 14: Why not use `Version` here? Honest question

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -225,6 +225,10 @@ void HexagonGame::initLua_Utils()
         .doc(
             "Force-swaps (180 degrees) the player when invoked. If `$0` is "
             "`true`, the swap sound will be played.");
+
+    addLuaFn("u_getVersion", //
+        [this] { return Config::getVersionString(); })
+        .doc("Returns the string representing the current version of the game");
```
SuperV1234, line 231: Instead of this, how about the functions:

- `u_getVersionMajor`
- `u_getVersionMinor`
- `u_getVersionMicro`

which would return integers, and then from the Lua side instead of doing string manipulation we could just do some basic int comparisons? The Lua function could accept `(major, minor, micro)` as arguments

SuperV1234: We could keep this one as `u_getVersionString` in case anyone wants that

SuperV1234 (changes requested):
Some more questions

include/SSVOpenHexagon/Global/Config.hpp
```diff
@@ -19,6 +20,32 @@ class Trigger;
 
 namespace hg::Config
 {
+struct Version
```
SuperV1234, line 23: Please move this to a `Version.hpp`  header

Morxemplum:
Made the changes. Also do keep in mind the parsing now takes into account if the "version" field of a profile is using the legacy version. If it is using the legacy format it'll register as an invalid version and the game will load it as (-1, 0, 0). Hopefully after writing to the profile it will write the new format so this should only happen for one launch.

SuperV1234 (approved):
LGTM

PR #296, opened 2020-09-20
Press F6 to reload the entire selected pack

IvoryDuke:
As by the title, added a new function to reload all levels in the currently selected pack when F6 is pressed.

Extras:
- modified level reload function to accept new files to match the behavior of the pack reload. Also condensed the whole process in just one function;
- implemented a safer way to agglomerate key binds on boot if there are empty slots in between (also much better looking);
- removed hardcoded left and right. Most people will bind the right and left arrows to left and right actions anyway causing the execution of both the hardcoded action and the customizable one (for ex. causing to move two levels in the level selection menu);
- cleaned up a couple of leftover comments from HexagonGame.cpp.

### DIFF
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -261,8 +261,7 @@ private:
         return getCurrentMenu() != nullptr;
     }
 
-    void reloadLevelAssets();
-    void reloadPackAssets();
+    void reloadAssets(const bool pack);
     int noActions{0};
 
 public:
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -262,6 +262,7 @@ private:
     }
 
     void reloadLevelAssets();
+    void reloadPackAssets();
     int noActions{0};
 
 public:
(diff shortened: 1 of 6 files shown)


include/SSVOpenHexagon/Core/MenuGame.hpp
```diff
@@ -261,7 +261,7 @@ class MenuGame
         return getCurrentMenu() != nullptr;
     }
 
-    void reloadLevelAssets();
+    void reloadAssets(const bool pack);
```
SuperV1234, line 264: Can we rename `pack` to something more meaningful? E.g.`reloadEntirePack`.

src/SSVOpenHexagon/Global/Assets.cpp
```diff
@@ -379,122 +379,303 @@ const StyleData& HGAssets::getStyleData(
...
-    if(!ssvufs::Path{path}.exists<ssvufs::Type::Folder>())
+            auto it = styleDataMap.find(temp);
+            if(it == styleDataMap.end())
+            {
+                styleDataMap.emplace(temp, std::move(styleData));
+            }
+            else
+            {
+                it->second = styleData;
+            }
```
SuperV1234, line 431: I think this could just be:

```cpp
styleDataMap[temp] = std::move(styleData);
```

And similarly below.

SuperV1234 (changes requested):
Very minor

PR #297, opened 2020-09-21
Options menu items to reset configs to defaults

IvoryDuke:
"reset configs to default" in the main options menu resets all configs.
"reset binds to defaults" in the gameplay submenu only resets all the keyboard-mouse-joystick binds.

### DIFF
--- a/src/SSVOpenHexagon/Global/Config.cpp
+++ b/src/SSVOpenHexagon/Global/Config.cpp
@@ -23,6 +23,31 @@ using namespace ssvu::FileSystem;
 using namespace ssvuj;
 using namespace ssvu;
 
+#define X_BINDSLINKEDVALUES                                                \
+    X(joystickSelect, unsigned int, "j_select")                            \
+    X(joystickExit, unsigned int, "j_exit")                                \
+    X(joystickFocus, unsigned int, "j_focus")                              \
+    X(joystickSwap, unsigned int, "j_swap")                                \
+    X(joystickForceRestart, unsigned int, "j_force_restart")               \
+    X(joystickRestart, unsigned int, "j_restart")                          \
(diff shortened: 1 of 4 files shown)


Morxemplum:
Personally I think the default values should be hardcoded in, not be read from a separate file. If it is going to be read from a separate file how do we know that the user didn't tamper with the values from that file? How do we know that file isn't deleted? Try hardcoding in the values so that way we can guarantee no tampering can break this functionality.

IvoryDuke:
Sorry but I disagree, this way of thinking makes the existence of the default_config.json itself meaningless.
In this case, the entire game should be reworked to never use such file, which is not the point of this PR.

src/SSVOpenHexagon/Global/Config.cpp
```diff
@@ -162,7 +165,22 @@ static void syncAllToObj()
 #undef X
 }
 
+static void resetAllFromObj()
+{
+#define X(name, type, key) name().syncFrom(ssvuj::getFromFile("../misc/default_config.json"));
```
SuperV1234, line 170: In the deployed game, `default_config.json` is in the main folder, not under `../misc`. I think this needs to be changed to `./default_config.json` (and below as well). Can you test this by copy-pasting your build in your Steam Open Hexagon folder (and replacing the exe)?

SuperV1234 (changes requested):

src/SSVOpenHexagon/Global/Config.cpp
IvoryDuke: I tried using default_config.json in the github build but it would always crash.
I am going to check what happens with the Steam game.

SuperV1234 (approved):
LGTM

PR #298, opened 2020-09-21
Increased Dialog Box Customization

IvoryDuke:
```
void create(const std::string& output, const int charSize, const float mFrameSize, const int mDrawMode,
        const float mXPos = 0.f, const float mYPos = 0.f);
void create(const std::string& output, const int charSize, const float mFrameSize, const int mDrawMode,
        const KKey mKeyToClose, const float mXPos = 0.f, const float mYPos = 0.f);
void draw(const Color& txtColor, const Color& backdropColor);
```
- mDrawMode 1 (DBoxDraw::topLeft) will place the box at {xPos, yPos} from the top left corner;
- mDrawMode 2 (DBoxDraw::center) will place the box at {xPos, yPos} from the screen center;
- mDrawMode 3 (DBoxDraw::centerUpperHalf) is a specialized version for the level reload function.
NOTE: xPos positive values shift the box to the right, yPos positive values shift it toward the bottom.

Also:
- added customization for the thickness of the frame surrounding the box and the color of both text and backdrop;
- condensed the functions to block the menu input processing into a single one, and renamed variable "noActions" to a more clear "ignoreInputs".

DrawMode 1 with a small x and y offset.
Capture

DrawMode 2 perfectly centered
Capture2

### DIFF
--- a/include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
@@ -20,8 +20,8 @@ class StyleData;
 enum DBoxDraw
 {
     topLeft = 0,
-    centered,
-    centeredUpperHalf
+    center,
+    centerUpperHalf
 };
 
 class HexagonDialogBox
@@ -29,6 +29,7 @@ class HexagonDialogBox
 private:
     using KKey = sf::Keyboard::Key;
     using Color = sf::Color;
+    using Vector2 = sf::Vector2f;
 
     HGAssets& assets;
     ssvs::GameWindow& window;
@@ -53,19 +54,22 @@ private:
     KKey keyToClose{KKey::Unknown};
 
     void drawText(const Color& txtColor, const float xOffset, const float yOffset);
-    void drawCentered(const Color& txtColor, const Color& backdropColor);
-    void drawCenteredUpperHalf(const Color& txtColor, const Color& backdropColor);
+    void drawBox(const Color& frameColor, const float x1, const float x2,
+        const float y1, const float y2);
+    void drawCenter(const Color& txtColor, const Color& backdropColor);
+    void drawCenterUpperHalf(const Color& txtColor, const Color& backdropColor);
     void drawTopLeft(const Color& txtColor, const Color& backdropColor);
 
 public:
     HexagonDialogBox(HGAssets& mAssets, ssvs::GameWindow& window, StyleData& styleData);
 
-    void createDialogBox(const std::string& output, const int charSize,
+    void create(const std::string& output, const int charSize,
         const float mFrameSize, const int mDrawMode,
         const float xPos = 0.f, const float yPos = 0.f);
-    void createDialogBox(const std::string& output, const int charSize,
+    void create(const std::string& output, const int charSize,
         const float mFrameSize, const int mDrawMode,
-        const KKey mKeyToClose, const float mXPos = 0.f, const float mYPos = 0.f);
+        const KKey mKeyToClose, const float mXPos = 0.f,
+        const float mYPos = 0.f);
 
     void draw(const Color& txtColor, const Color& backdropColor);
     void clearDialogBox();
--- a/include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
@@ -17,6 +17,13 @@ namespace hg
 class HGAssets;
 class StyleData;
 
+enum DBoxDraw
+{
+    regular = 0,
+    centered,
+    levelReload
+};
+
 class HexagonDialogBox
 {
 private:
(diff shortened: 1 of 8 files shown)


Morxemplum:
I feel like in addition there should be a way where we can customize whether we can use "any key" or only using the "select" key to be able to close it. There are certain situations where "select" key would be more appropriate to use than "any key". If you can get this down then I would personally let this one pass through.

IvoryDuke:
Done, here is an example of the necessary procedure
```
// code where you edit the message to ask user to press O to close the box

setIgnoreInputs(1); // make sure menu does not activate regular actions
dBoxClosingKey = KKey::O; // assign the closing key
dialogBox.createDialogBox(reloadOutput, 26, 10.f, DBoxDraw::centered); // create the box
```

Morxemplum:
This solution works for now. When the UI starts to come around we will have to figure out how to actually implement buttons to make it more mouse friendly.

include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
```diff
@@ -17,31 +17,67 @@ namespace hg
 class HGAssets;
 class StyleData;
 
+enum DBoxDraw
+{
+    topLeft = 0,
+    center,
+    centerUpperHalf
+};
```
SuperV1234, line 25: Please change this to an `enum class` so that the enumerator names do not leak in the surrounding namespace. This will require you explicitly saying `DBoxDraw::topLeft` rather than just `topLeft` on the usage site, which is a good thing for readability and avoiding name collisions. 

```suggestion
enum class DBoxDraw
{
    topLeft = 0,
    center,
    centerUpperHalf
};
```

include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
```diff
@@ -17,31 +17,67 @@ namespace hg
...
+
 class HexagonDialogBox
 {
 private:
+    using KKey = sf::Keyboard::Key;
+    using Color = sf::Color;
+    using Vector2 = sf::Vector2f;
```
SuperV1234, line 32: Mostly personal style but I've grown to avoid aliases whose only purpose is to make code shorter. Up to you, not asking you to remove them.

include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
```diff
@@ -17,31 +17,67 @@ namespace hg
...
 
-    void createDialogBox(const std::string& output, const int charSize);
-    void drawDialogBox();
+    void draw(const Color& txtColor, const Color& backdropColor);
     void clearDialogBox();
 
+    [[nodiscard]] KKey getKeyToClose() const noexcept
+    {
+        return keyToClose;
+    }
```
SuperV1234, line 80:
```suggestion
    }
    
```

include/SSVOpenHexagon/Core/MenuGame.hpp
```diff
@@ -262,7 +262,9 @@ class MenuGame
     }
 
     void reloadLevelAssets();
-    int noActions{0};
+
+    void setIgnoreInputs(const int keyPresses);
+    int ignoreInputs{0};
```
SuperV1234, line 267: This is a much better name, +1

include/SSVOpenHexagon/Core/HexagonDialogBox.hpp
```diff
@@ -17,31 +17,67 @@ namespace hg
...
+    using KKey = sf::Keyboard::Key;
+    using Color = sf::Color;
+    using Vector2 = sf::Vector2f;
+
     HGAssets& assets;
     ssvs::GameWindow& window;
     StyleData& styleData;
     sf::Font& imagine;
 
+    std::function<void(const Color&, const Color&)> drawFunc;
```
SuperV1234, line 39: This is one of the places where I would use an alias, as I suggest some refactoring below.

```suggestion
    using DrawFunc = std::function<void(const Color&, const Color&)>;
    DrawFunc drawFunc;
```

src/SSVOpenHexagon/Core/HexagonDialogBox.cpp
```diff
@@ -12,27 +12,48 @@ namespace hg
...
+            drawCenter(txtColor, backdropColor);
+        };
+        break;
+
+    case DBoxDraw::centerUpperHalf:
+        drawFunc = [this](const Color& txtColor, const Color& backdropColor) {
+          drawCenterUpperHalf(txtColor, backdropColor);
+        };
+        break;
+    }
```
SuperV1234, line 53: I would extract this bit to a private member function:

```cpp
[[nodiscard]] DrawFunc drawModeToDrawFunc(DBoxDraw drawMode) const
{
    switch(drawMode)
    {
    case DBoxDraw::topLeft:
        return [this](const Color& txtColor, const Color& backdropColor) {
            drawTopLeft(txtColor, backdropColor);
        };

    case DBoxDraw::center:
        return [this](const Color& txtColor, const Color& backdropColor) {
            drawCenter(txtColor, backdropColor);
        };
    
    default:
        assert(drawMode == DBoxDraw::centerUpperHalf);
        return [this](const Color& txtColor, const Color& backdropColor) {
            drawCenterUpperHalf(txtColor, backdropColor);
        };
    }
}
```

src/SSVOpenHexagon/Core/HexagonDialogBox.cpp
```diff
@@ -49,16 +70,49 @@ void HexagonDialogBox::createDialogBox(
...
 
-    const float w = Config::getWidth() * fmax;
-    const float h = (Config::getHeight() * fmax) / 2.f;
+void HexagonDialogBox::drawBox(const Color& frameColor, const float x1,
+    const float x2, const float y1, const float y2)
+{
+    sf::Vector2f p1{x1, y1}; // top left
+    sf::Vector2f p2{x2, y1}; // top right
+    sf::Vector2f p3{x2, y2}; // bottom right
+    sf::Vector2f p4{x1, y2}; // bottom left
```
SuperV1234, line 92:
```suggestion
    sf::Vector2f topLeft{x1, y1};
    sf::Vector2f topRight{x2, y1};
    sf::Vector2f bottomRight{x2, y2};
    sf::Vector2f bottomLeft{x1, y2};
```

SuperV1234 (changes requested):
Very good improvements overall. Some minor things to change

SuperV1234 (approved):
LGTM

PR #300, opened 2020-10-08
[Incompatibility] Move message functions to Event Timeline Prefix. Create New Prefix a_ for Audio Handling.

Morxemplum:
This is a PR that will introduce incompatibilities to the game, but it's best that we do this now rather than later where this can hurt us more down the line.

Essentially the message functions that we all know will be removed from the **m_** prefix and will merge with the **e_** prefix. It makes it more intuitive for new developers to understand that messages run on the event timeline.

In addition, I've also decided to move some of the utility functions that specifically control audio and I decided to move them to their own prefix, **a_**. This is being done to remove some of the bloating of the utilities prefix. This also introduces an incompatibility, but it shouldn't be as significant as the messages because the audio functions aren't used as frequently. The main functions that are being moved to **a_** are the music functions and the sound functions **u_playSound** and **u_playPackSound**. It's also a good idea to have a prefix dedicated to audio functions because I plan on adding several functions that would fit perfectly under this prefix.

I also made a few smaller changes in this PR as well:

- When the "color" attribute of a Style is not provided in the JSON, it will default to the color black instead of white. This technically introduces an incompatibility, but since every style explicitly defines the color and pulse attributes, this incompatibility shouldn't break anything and it is negligible. This change is mainly done to allow people to shortcut certain colors and panels.
- Slightly touched up the documentation for not only the new prefix, but to also update a few things due to new functions coming to 2.0.3

### DIFF
--- a/_RELEASE/Packs/cube/Scripts/Levels/apeirogon.lua
+++ b/_RELEASE/Packs/cube/Scripts/Levels/apeirogon.lua
@@ -61,7 +61,7 @@ end
 -- onLoad is an hardcoded function that is called when the level is started/restarted
 function onLoad()
 	if (u_getDifficultyMult() >= 1.25) then
-		m_messageAdd("Difficulty >= 1.25\nPentagon removed!", 120)
+		e_messageAdd("Difficulty >= 1.25\nPentagon removed!", 120)
 		l_setSidesMin(6)
 	end
 end
@@ -82,7 +82,7 @@ end
 function onIncrement()
 	enableSwapIfSpeedGEThan(4);
 	if (u_getSpeedMultDM() >= 4.5 and l_getSidesMin() == 5) then
-		m_messageAddImportant("Speed >= 4.5\nPentagon removed!", 120)
+		e_messageAddImportant("Speed >= 4.5\nPentagon removed!", 120)
 		if (l_getSides() == 5) then
 			l_setSides(6)
 		end
--- a/_RELEASE/Packs/cube/Scripts/Levels/babysteps.lua
+++ b/_RELEASE/Packs/cube/Scripts/Levels/babysteps.lua
@@ -38,21 +38,21 @@ end
 
 -- onLoad is an hardcoded function that is called when the level is started/restarted
 function onLoad()
-	m_messageAddImportant("welcome to open hexagon 2", 130)
-	m_messageAddImportant("use left/right to rotate", 130)
-	m_messageAddImportant("avoid the walls!", 130)
+	e_messageAddImportant("welcome to open hexagon 2", 130)
+	e_messageAddImportant("use left/right to rotate", 130)
+	e_messageAddImportant("avoid the walls!", 130)
 	e_eventStopTimeS(6)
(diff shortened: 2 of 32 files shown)


Morxemplum:
Just to keep in mind, this incompatibility is NOT hard to fix at all. Simple Find and Replace operations can easily fix these incompatibilities. I will be glad to make a video tutorial on how to fix these incompatibilities so packs can work in 2.0.3

SuperV1234:
I understand where you're coming from and I agree with your overall vision and plan, but I do not want to introduce incompatibilities and I want to minimize breakage of existing packs, even if it's easy to fix for level developers. What I would suggest is to keep both `m_messageAdd' and `e_messageAdd`, and either (1) explicitly mention that the former is deprecated and eligible for removal in the future, or (2) hide `m_messageAdd` from the documentation and pretend it does not exist.

SuperV1234 (changes requested):
See comment above.

Morxemplum:
Alright finally got time to do what you asked for. So now I brought back the **m_** prefix for the message functions, because I can totally understand where you're coming from when you say you want to minimize the incompatibilities. The message functions are used so frequently that removing it immediately will cause a majority of pack levels to break, which is what we absolutely don't want.

However, all of the functions that are migrated over to the audio prefix will still be removed immediately. The amount of levels breaking from this incompatibility will be nowhere near as bad as the message prefix incompatibility.

Alongside bringing back **m_** temporarily, I have also implemented a warning system into the game for these deprecated functions, so for every first call of a deprecated function in a level session, a warning is raised to the console telling them the function is deprecated and they should be updating it. The in level GUI has been updated to show the amount of warnings raised in the session if at least 1 warning has been raised. 

Now of course the warning message comes off as a bit annoying, and that's what we want. With this annoying warning message, it should create incentives for pack developers to fix the warnings and update their packs so they will not break if the deprecated functions eventually get removed. Now you probably will want me to add a "Suppress Warnings" option to not have the message display, but I don't want that to be added right now. I really want to make sure that people actually update their packs to fix the warnings at the moment so their packs don't break when those deprecated functions get removed. I will be glad to add a "Suppress Warnings" option post-release of the game, because we won't be able to easily remove deprecated functions like this post-release.

But these commits should hopefully address your concerns. Below is an image of what the warning message looks like and I also show what the warning message in the console is.

**[Lua] WARNING: The function "m_messageAddImportant" (used in level "lasers") is deprecated. This function will be removed in a future version of Open Hexagon. Please replace all occurrences of this function with "e_messageAddImportant" in your level files and common.lua.**

SSVOpenHexagon_2020-10-24_16-18-35

Morxemplum:
In addition, with the new **a_** prefix, I have many ideas for new functions to add to that prefix, such as pitch control and asset overriding. Hopefully I can implement these quick enough so they can make it into 2.0.3 (as I would say the full release is coming soon). Also forgot to request review so I'm doing it now

SuperV1234 (approved):
LGTM, will make some last adjustments myself

PR #305, opened 2020-11-11
Further menu improvements

IvoryDuke:
- Fixed pack switch still getting stuck in some circumstances I cannot figure out (assignement is now constantly updated instead of being bound to key press/release events);

- Added wrapping to the level description. It would be better if pack makers updated their jsons with newlines to fit the space available as they see fit, but since it is possible some packs may never be updated I enforce a formatting when the level is changed.
It should be noted that I have noticed some weird glitches in the formatted text rendering in certain packs (particularly Babagon). Best I can tell it's due to the editor used to make the jsons since if I retype the description as in the original in my text editors the issue is fixed.

### DIFF
--- a/src/SSVOpenHexagon/Core/MenuGame.cpp
+++ b/src/SSVOpenHexagon/Core/MenuGame.cpp
@@ -130,21 +130,9 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
         }
     };
 
-    game.onEvent(Event::EventType::KeyPressed) += [this](const Event& mEvent) {
-        if(mEvent.key.code == KKey::LShift || mEvent.key.code == KKey::RShift)
-        {
-            focusHeld = true;
-        }
-    };
-
     game.onEvent(Event::EventType::KeyReleased) += [this, checkCloseBootScreens,
                                                        checkCloseDialogBox](
                                                        const Event& mEvent) {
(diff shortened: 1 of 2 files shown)


src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -1834,6 +1820,9 @@ void MenuGame::update(ssvu::FT mFT)
         case States::LoadingScreen: hexagonRotation += mFT / 100.f; return;
 
         case States::LevelSelection:
+            focusHeld = sf::Keyboard::isKeyPressed(KKey::LShift) ||
+                        sf::Keyboard::isKeyPressed(KKey::RShift);
```
SuperV1234, line 1824: I'm OK with this as a first version, but focus can be rebound, though. We should probably get the binding from the config.

SuperV1234 (approved):

src/SSVOpenHexagon/Core/MenuGame.cpp
IvoryDuke: Makes sense, I will take care of it next week.

IvoryDuke: Would it be ok if I used the Focus bind for this?
```
add2StateInput(game, Config::getTriggerFocus(), focusHeld, Tid::Focus);
```

PR #306, opened 2020-11-11
Customizable Quick Pack Switch key

IvoryDuke:
Assigned Focus bind to control the quick switch function when in menu (its in game behavior is unchanged).
Also:
- replaced a #define in Config.cpp with a constexpr;
- added pack change keys to RefreshBinds();
- small tweak in formatLevelDescription().

A new custom bind can be made but that would make the bind customization menu go out of bounds and would require a bigger amount of effort.

### DIFF
--- a/src/SSVOpenHexagon/Core/Joystick.cpp
+++ b/src/SSVOpenHexagon/Core/Joystick.cpp
@@ -134,6 +134,11 @@ enum class AxisDir : int
     Right = 1
 };
 
+inline AxisDir operator-(AxisDir dir)
+{
+    return static_cast<AxisDir>(- static_cast<int>(dir));
+}
+
 [[nodiscard]] static AxisDir axisPressed(
     const unsigned int joyId, const sf::Joystick::Axis axis)
 {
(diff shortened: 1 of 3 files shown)


IvoryDuke:
Working on making it work for controller too.

src/SSVOpenHexagon/Core/Joystick.cpp
```diff
@@ -134,6 +134,11 @@ enum class AxisDir : int
     Right = 1
 };
 
+inline AxisDir operator-(AxisDir dir)
```
SuperV1234, line 137:
```suggestion
[[nodiscard]] AxisDir operator-(AxisDir dir)
```

No need for `inline` here, TBH.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -1686,6 +1688,16 @@ void MenuGame::update(ssvu::FT mFT)
...
 
+    if(hg::Joystick::nextPackRisingEdge())
+    {
+        changePackAction(1);
+    }
+    else if(hg::Joystick::previousPackRisingEdge())
+    {
+        changePackAction(-1);
+    }
+    focusHeld = hg::Joystick::focusPressed();
```
SuperV1234, line 1699:
```suggestion
    }
    
    focusHeld = hg::Joystick::focusPressed();
```

I like vertical space :)

SuperV1234 (commented):
Looks good so far. Waiting for the controller changes before merging it in!

IvoryDuke:
In case it was not clear: controller is fully working

SuperV1234 (approved):

PR #307, opened 2020-11-24
Expand the audio prefix (First Expansion)

### LINKED ISSUE #277: [Could Have] Add custom swap sounds level-by-level
In 1.92, custom swap systems had a creative outlet by allowing pack developers to make their own sound with the swap. This technically is possible with 2.x's onInput function, but that means you wouldn't be able to have the "Swap Enabled" on the top left and there would be no color to enable for swapping (unless you do it manually with style switching). Perhaps there could be a lua function that can be added to override the programmer swap sound with a custom one in the "Sounds" folder.

Morxemplum:
This PR introduces 6 new Lua functions that all expand the new **a_** prefix that was introduced.

Two of these functions are dedicated to the pitch of the music

1. **a_syncMusicToDM**: Allows pack developers to override the user's preference of whether the music's pitch is synced with the difficulty multiplier. By default, the value is based on the user's preference.

2. **a_setMusicPitch**: Allows pack developers to manually adjust the pitch of the music playing. Due to SFML limitations, this function does not accept negative values.

The other four are asset overriding functions, which allow pack developers to override some of the default sounds with their own custom ones.

1. **a_overrideBeepSound**: Overrides beep.ogg with a level pack sound to the developer's liking.
2. **a_overrideIncrementSound**: Overrides increment.ogg with a level pack sound to the developer's liking.
3. **a_overrideSwapSound**: Overrides swap.ogg with a level pack sound to the developer's liking. Closes #277.
4. **a_overrideDeathSound**: Overrides death.ogg with a level pack sound to the developer's liking.

I'm aware several more audio-related functions can be added later down the line, but these are a few starters.

### DIFF
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -474,8 +474,14 @@ public:
 
     [[nodiscard]] float getMusicDMSyncFactor() const
     {
-        return Config::getMusicSpeedDMSync() ? std::pow(difficultyMult, 0.12f)
-                                             : 1.f;
+        return levelStatus.syncMusicToDM ? std::pow(difficultyMult, 0.12f)
+                                         : 1.f;
+    }
+
+    void setMusicPitch(sf::Music* current)
+    {
+        current->setPitch((getMusicDMSyncFactor()) *
+                          Config::getMusicSpeedMult() * levelStatus.musicPitch);
     }
 
     // Input
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -478,7 +478,7 @@ public:
                                          : 1.f;
     }
 
-    void setMusicPitch(sf::Music* current)
+    void setMusicPitch(sf::Music& current)
     {
         current->setPitch((getMusicDMSyncFactor()) *
                           Config::getMusicSpeedMult() * levelStatus.musicPitch);
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -480,7 +480,7 @@ public:
 
     void setMusicPitch(sf::Music& current)
     {
-        current->setPitch((getMusicDMSyncFactor()) *
+        current.setPitch((getMusicDMSyncFactor()) *
                           Config::getMusicSpeedMult() * levelStatus.musicPitch);
     }
 
--- a/src/SSVOpenHexagon/Core/HGScripting.cpp
+++ b/src/SSVOpenHexagon/Core/HGScripting.cpp
@@ -247,6 +247,32 @@ void HexagonGame::initLua_AudioControl()
         .doc(
             "Dives into the `Sounds` folder of the current level pack and "
             "plays the specified file `$0`.");
+
+    addLuaFn("a_syncMusicToDM", //
+        [this](bool value) {
+            levelStatus.syncMusicToDM = value;
+
+            auto* current(assets.getMusicPlayer().getCurrent());
+            setMusicPitch(current);
+        })
+        .arg("value")
+        .doc(
+            "This function, when called, overrides the user's preference of "
+            "adjusting the music's pitch to the difficulty multiplier. Useful "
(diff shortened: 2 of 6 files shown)


include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -474,8 +474,14 @@ class HexagonGame
 
     [[nodiscard]] float getMusicDMSyncFactor() const
     {
-        return Config::getMusicSpeedDMSync() ? std::pow(difficultyMult, 0.12f)
-                                             : 1.f;
+        return levelStatus.syncMusicToDM ? std::pow(difficultyMult, 0.12f)
+                                         : 1.f;
+    }
+
+    void setMusicPitch(sf::Music* current)
```
SuperV1234, line 481: If you are expecting `current` to never be `nullptr`, it is recommended to accept it by reference to avoid introducing the extra null state in the type system:

```suggestion
    void setMusicPitch(sf::Music& current)
```

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -474,8 +474,14 @@ class HexagonGame
...
+    {
+        current->setPitch((getMusicDMSyncFactor()) *
+                          Config::getMusicSpeedMult() * levelStatus.musicPitch);
```
SuperV1234, line 484:
```suggestion
        current.setPitch((getMusicDMSyncFactor()) *
                          Config::getMusicSpeedMult() * levelStatus.musicPitch);
```

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -247,6 +247,74 @@ void HexagonGame::initLua_AudioControl()
         .doc(
             "Dives into the `Sounds` folder of the current level pack and "
             "plays the specified file `$0`.");
+
+    addLuaFn("a_syncMusicToDM", //
+        [this](bool value) {
+            levelStatus.syncMusicToDM = value;
+
+            auto* current(assets.getMusicPlayer().getCurrent());
+            setMusicPitch(current);
```
SuperV1234, line 256: Not sure if this can happen, but let's protect against undefined behavior:

```suggestion
            sf::Music* current(assets.getMusicPlayer().getCurrent());
            if (current == nullptr)
            {
                return:
            }
            
            setMusicPitch(*current);
```

SuperV1234, line 256: Shouldn't this change `levelStatus.syncMusicToDM` instead of calling `setMusicPitch`?

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -247,6 +247,74 @@ void HexagonGame::initLua_AudioControl()
...
+            "This function, when called, overrides the user's preference of "
+            "adjusting the music's pitch to the difficulty multiplier. Useful "
+            "for levels that rely on the music to time events.");
+
+    addLuaFn("a_setMusicPitch", //
+        [this](float mPitch) {
+            levelStatus.musicPitch = mPitch;
+
+            auto* current(assets.getMusicPlayer().getCurrent());
+            setMusicPitch(current);
```
SuperV1234, line 269: Ditto:

```suggestion
            sf::Music* current(assets.getMusicPlayer().getCurrent());
            if (current == nullptr)
            {
                return:
            }
```

SuperV1234 (changes requested):
I think there's a bug, and a few possible improvements. Also, regarding the four overriding functions -- have you tested them? I am not sure what the path should look like, and I don't think it's going to work this way.

If you call `a_overrideBeepSound("beep.ogg")`, it will look into `./Assets/ right? I would like it if we could look into the pack folder as well. Maybe introduce some sort of magic variable like:

```cpp
a_overrideBeepSound("${PACK_FOLDER}/mysound.ogg")
```

?

include/SSVOpenHexagon/Core/HexagonGame.hpp
Morxemplum: Ah, I see. This was the original case of how the auto was defined in a code block, and I guess this was a mistake in the code originally. I'll commit to your suggestion.

SuperV1234 (approved):
LGTM

PR #309, opened 2020-11-28
My 2.0.3 Wrap Up Changes

Morxemplum:
This PR contains all of the last minute changes that I want to make to the game before a 2.0.3 release. I don't plan on working on any more changes for 2.0.3 after this PR unless it's bug fixes.

Many things go on in this PR, so here's a quick rundown of all the changes I've made.

1. Custom Walls have a new attribute: Collision. It was a desperately requested feature, and it was simpler than I thought to get it implemented. When the collision is set to false, not only do the walls not kill the player, but the player can also move through the walls without a problem. This allows custom walls to be used for other purposes.

2. I rewrote the Discord Rich Presence. I found the current implementation of Discord's Rich Presence to be bulky and unfun, and it looked gross. So I went through the painstaking effort to learn the Discord extlib a bit and rewrote the Rich Presence to be snappier and present the information more cleanly. I take advantage of Discord's Timestamps with the Rich Presence and use them to "track" a person's current attempt in real-time. Also, I show the pack that the level comes from, the personal best they got with the level and whether they are watching a replay. Below are some images of what it looks like. Now, a few improvements can be made, such as improving the replay information to show what they are watching and even using some visual aid to show what the game can look like to those who've never heard of the game. Overall, I'm much more satisfied with this Rich Presence.
Discord_2020-11-27_02-09-24
AsPowerBar_2020-11-28_00-28-22

3. The distance at which standard walls spawn can now be customized through **l_setWallSpawnDistance**. Also, the default wall spawn distance has been lowered to 1300. This only applies to walls that don't accelerate (walls that have acceleration still spawn at 1600 for compatibility reasons). This can help assist in pattern design, can help be used to deliver walls more quickly, or hide the spawning visual with more zoomed out levels.

4. Closely related to number 3, I've added out of bounds checking for standard walls, for those pesky wallAccs that decide to go backward and you need to increment. The outer bounds are determined by the wall spawn distance, which means the aforementioned Lua function can change it.

5. I've asked Lukemans to design two new sounds for the difficulty multiplier. With the current UI as is, it's hard to distinguish when you change the difficulty. So I decided to make use of two new sounds to help make some distinguishment. I'm assuming some visual aid will be implemented as well, but this is a good start.

In this PR, a few bugs are now fixed:

1. I've fixed the collision issue experienced after dying in Tutorial Mode. It was a simple fix.

2. When Rich Presence fails callbacks more than 20 times, Rich Presence stops getting updated altogether. Now the console can stop getting spammed, and we can look at what's important.

3. The death ring no longer lingers after you hit a wall with invincibility. In both Tutorial Mode and Invincibility Mode, the death animation ends after a certain point.
     a. Also, I also fixed a small bug where the animation doesn't play when swap colors are happening.

Along with this PR, I've gone out of my way and deprecated several more functions that just didn't make sense. It's important that new pack developers can easily memorize and understand Lua functions, and it's best we deprecate them now and give current pack developers more time to adjust. Here are all of the functions now deprecated.

- **u_kill** is now **t_kill**. This is because the killing happens on the pattern timeline, so it should be in the t_ prefix.
- **u_eventKill** is now **e_kill**. Same reasoning, but on the event timeline.
- **e_eventStopTime** is now **e_stopTime**.
- **e_eventStopTimeS** is now **e_stopTimeS**.
- **e_eventWait** is now **e_wait**.
- **e_eventWaitS** is now **e_waitS**.
- **e_eventWaitUntilS** is now **e_waitUntilS**.

Yeah. That's a lot of changes, but these are important for players and pack developers alike, and it's best they are done before a release rather than be stuck in beta. Anyway, that's all I have! If there are other last-minute changes, I suggest someone either makes their own PR (or if you're Vee, just commit to master directly).


SuperV1234:
> the default wall spawn distance has been lowered to 1300

Why?

> walls that have acceleration still spawn at 1600 for compatibility reasons

Doesn't this break existing patterns that are a mix of normal and accelerating walls?

include/SSVOpenHexagon/Components/CCustomWall.hpp
```diff
@@ -27,6 +27,8 @@ class CCustomWall
 private:
     std::array<sf::Vector2f, 4> vertexPositions;
     std::array<sf::Color, 4> vertexColors;
+    bool canCollide{true};
+    int8_t renderOrder{1};
```
SuperV1234, line 31: I'm assuming this is not used at the moment -- please comment it out so that it doesn't take additional memory until we figure out how to do this properly.

include/SSVOpenHexagon/Components/CCustomWall.hpp
```diff
@@ -52,11 +54,26 @@ class CCustomWall
...
 
+    [[gnu::always_inline]] void setCanCollide(const bool collide) noexcept
+    {
+        canCollide = collide;
+    }
+
+    [[gnu::always_inline]] void setRenderOrder(const int8_t order) noexcept
+    {
+        renderOrder = order;
+    }
```
SuperV1234, line 65: Ditto

include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -63,6 +63,11 @@ class HexagonGame
     ssvs::GameWindow& window;
 
 public:
+    bool discordHung{false};
+    bool steamHung{false};
+    int8_t discordAttempt{1};
+    int8_t steamAttempt{1};
```
SuperV1234, line 69:
```suggestion
    std::int8_t discordAttempt{1};
    std::int8_t steamAttempt{1};
```

SuperV1234: Also please include the `<cstdint>` header, that's where `std::int8_t` is defined.

include/SSVOpenHexagon/Data/LevelData.hpp
```diff
@@ -106,6 +107,9 @@ struct LevelStatus
     float wallSkewRight{0.f};
     float wallAngleLeft{0.f};
     float wallAngleRight{0.f};
+    float wallSpawnDistance{1350.f};
+    float wallAccSpawnDistance{
+        Config::getSpawnDistance()}; // Attribute for compatibility
```
SuperV1234, line 112: This needs some more explanation/context as I mentioned in the PR comment

src/SSVOpenHexagon/Components/CWall.cpp
```diff
@@ -55,23 +55,34 @@ void CWall::update(HexagonGame& mHexagonGame, ssvu::FT mFT)
...
             ++pointsOnCenter;
-        }
+        } 
         else
         {
+			if(std::abs(vp.x - mCenterPos.x) > outerBounds &&
+                std::abs(vp.y - mCenterPos.y) > outerBounds)
+			{
+				++pointsOutOfBounds;
+			}
```
SuperV1234, line 80: Why was this necessary? When do normal walls or accelerating walls end up out of bounds? Maybe a timer approach would be cheaper compared to an extra collision check per frame?

src/SSVOpenHexagon/Components/CWall.cpp
```diff
@@ -55,23 +55,34 @@ void CWall::update(HexagonGame& mHexagonGame, ssvu::FT mFT)
...
             ssvs::moveTowards(vp, mCenterPos, speed.speed * 5.f * mFT);
         }
     }
 
-    if(pointsOnCenter > 3)
+    if((pointsOnCenter > 3 || pointsOutOfBounds > 3) || (pointsOnCenter == 2 && pointsOutOfBounds == 2))
```
SuperV1234, line 85: This needs a comment in the code, not sure why it is the way it is

src/SSVOpenHexagon/Components/CWall.cpp
```diff
@@ -55,23 +55,34 @@ void CWall::update(HexagonGame& mHexagonGame, ssvu::FT mFT)
 void CWall::moveTowardsCenter(
     HexagonGame& mHexagonGame, const sf::Vector2f& mCenterPos, ssvu::FT mFT)
 {
+    const float wallSpawnDist{mHexagonGame.getLevelStatus().wallSpawnDistance};
+    const float wallAccSpawnDist{mHexagonGame.getLevelStatus().wallAccSpawnDistance};
+
     const float radius{mHexagonGame.getRadius() * 0.65f};
+    const float outerBounds =
+        (wallSpawnDist == wallAccSpawnDist) ? wallSpawnDist * 0.5f : 800.f;
```
SuperV1234, line 63: Please add a comment here, why the hardcoded `800.f` and comparison between the two distances?

src/SSVOpenHexagon/Core/CCustomWallManager.cpp
```diff
@@ -94,6 +94,42 @@ void CCustomWallManager::setVertexPos(const CCustomWallHandle cwHandle,
...
+            << cwHandle << '\n';
+
+        SSVU_ASSERT(ssvu::contains(_freeHandles, cwHandle));
+        return;
+    }
+
+    SSVU_ASSERT(isValidHandle(cwHandle));
+
+    _customWalls[cwHandle].setRenderOrder(order);
+}
```
SuperV1234, line 131: Ditto, please comment out if unused.

src/SSVOpenHexagon/Core/Discord.cpp
```diff
@@ -89,32 +89,75 @@ bool discord_manager::set_rich_presence_in_menu()
...
-    activity.SetState("In game");
-    activity.SetDetails(str_status.data());
+    activity.SetDetails(("Playing " + level_info).data());
+    activity.SetState(second_info.data());
+    discord::ActivityTimestamps& currentTimestamp = activity.GetTimestamps();
+    if(!dead)
+    {
+        // ssvu::lo("Debug") << "The current time: " <<
+        // static_cast<int64_t>(std::chrono::high_resolution_clock::now().time_since_epoch().count()
+        // / 1000000000) << "\n";
```
SuperV1234, line 146: Please remove this if it was just for debugging

src/SSVOpenHexagon/Core/Discord.cpp
```diff
@@ -89,32 +89,75 @@ bool discord_manager::set_rich_presence_in_menu()
...
 {
     if(!_initialized)
     {
         return false;
     }
 
     discord::Activity activity{};
-    activity.SetState("In game");
-    activity.SetDetails(str_status.data());
+    activity.SetDetails(("Playing " + level_info).data());
```
SuperV1234, line 139: Is this well-defined behavior? You're creating a string temporary, giving its pointer to `SetDetails`, and then it dies afterwards. If `SetDetails` copies the characters then it's fine, but it just keeps a reference/pointer then it's UB. We should make sure and add a comment in the code if it's safe.

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -50,20 +31,44 @@ void HexagonGame::update(ssvu::FT mFT)
...
     {
         steamManager.set_rich_presence_in_game(nameStr, diffStr, timeStr);
-        discordManager.set_rich_presence_in_game(presenceStr);
+
         timeUntilRichPresenceUpdate = DELAY_TO_UPDATE;
     }
 
-    steamManager.run_callbacks();
-    discordManager.run_callbacks();
+    // Update Steam Rich Presence
```
SuperV1234, line 41: These two blocks of code warrant their own `private` member functions.

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -50,20 +31,44 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    if(!steamHung)
+    {
+        if(!steamManager.run_callbacks())
+        {
+            steamAttempt += 1;
+            if(steamAttempt > 20)
+            {
+                steamHung = true;
+                ssvu::lo("Steam") << "Too many failed callbacks. Stopping "
+                                     "Rich Presence updates."
```
SuperV1234, line 51: Is "Rich Presence" here accurate? I think we do more with steam callbacks

src/SSVOpenHexagon/Core/HGUpdate.cpp
```diff
@@ -50,20 +31,44 @@ void HexagonGame::update(ssvu::FT mFT)
...
+    {
+        if(!discordManager.run_callbacks())
+        {
+            discordAttempt += 1;
+            if(discordAttempt > 20)
+            {
+                discordHung = true;
+                ssvu::lo("Discord") << "Too many failed callbacks. Stopping "
+                                       "Rich Presence updates."
+                                    << "\n";
```
SuperV1234, line 68:
```suggestion
                ssvu::lo("Discord") << "Too many failed callbacks. Stopping "
                                       "Rich Presence updates.\n";
```

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -47,8 +47,11 @@ namespace
 void HexagonGame::createWall(int mSide, float mThickness,
     const SpeedData& mSpeed, const SpeedData& mCurve, float mHueMod)
 {
-    walls.emplace_back(*this, centerPos, mSide, mThickness,
-        Config::getSpawnDistance(), mSpeed, mCurve);
+    float spawnDistance = (mSpeed.accel == 0.f)
```
SuperV1234, line 50:
```suggestion
    const float spawnDistance = (mSpeed.accel == 0.f)
```

src/SSVOpenHexagon/Core/HexagonGame.cpp
```diff
@@ -473,6 +476,26 @@ void HexagonGame::death(bool mForce)
     assets.playSound("gameOver.ogg", ssvs::SoundPlayer::Mode::Abort);
     runLuaFunctionIfExists<void>("onDeath");
 
+    std::string nameStr = levelData->name;
```
SuperV1234, line 479: Bring this to down to line 492, and leave an empty line between the PB stuff and the rest. Easier to group visually

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -1762,8 +1767,37 @@ void MenuGame::exitAction()
...
+            hexagonGame.discordAttempt += 1;
+            if(hexagonGame.discordAttempt > 20)
+            {
+                hexagonGame.discordHung = true;
+                ssvu::lo("Discord") << "Too many failed callbacks. Stopping "
+                                       "Rich Presence updates."
+                                    << "\n";
+            }
+        }
+    }
```
SuperV1234, line 1800: Code repetition with `HexagonGame`. This should be extracted to a separate function (or two) in a separate file, and then both `MenuGame` and `HexagonGame` can depend on it and invoke it.

SuperV1234 (changes requested):
Generally good. I need more context on the spawn distance changes, and also a few fixes/improvements in terms of code quality

include/SSVOpenHexagon/Components/CCustomWall.hpp
Morxemplum: Like I mentioned in the original post, when wallAccs decelerate and start traveling away from the center, they never get deleted and as a result, this is why pack developers couldn't rely on the native method of incrementing to increment their levels. (Trust me, you'll see why I did this when I start messing with acceleradiant).

A timer would be way more difficult to implement. How would we know how long it takes for the wall to travel to the center (even if we account a change in speed)? How do we ensure the wall doesn't get terminated too early? I feel like trying to make all of those calculations will cost more performance than the current implementation. I think the current implementation is better (but I will weed out some calculations because I'm reverting some changes).

Morxemplum: Oh ffs ofc this replied to the wrong comment. Thanks GitHub mobile!

src/SSVOpenHexagon/Core/Discord.cpp
InfoTeddy: Here is what `SetDetails` does:
```c++
void Activity::SetDetails(char const* details)
{
    strncpy(internal_.details, details, 128);
    internal_.details[128 - 1] = '\0';
}
```
So it just copies the string over. And a comment should be added in the code to clarify.

Morxemplum: I'm also wondering if he wants me to remove "Playing " which I am completely fine with. It's not necessary information.

Morxemplum:
> > the default wall spawn distance has been lowered to 1300
> 
> Why?
> 
> > walls that have acceleration still spawn at 1600 for compatibility reasons
> 
> Doesn't this break existing patterns that are a mix of normal and accelerating walls?

Fair point. I didn't take this into consideration. Alright, I guess I can revert that. Would also help simplify other areas of code.

src/SSVOpenHexagon/Core/HGUpdate.cpp
Morxemplum: I will make the callbacks their own function, but it has to be public because MenuGame has to run this function as well.

src/SSVOpenHexagon/Components/CWall.cpp
Morxemplum: The first half of the condition will destroy the wall if it is in the center or out of bounds. The second half addresses an edge case where a wall is so thick it can reach the center and out of bounds.

Morxemplum:
A lot of the pull requests I'm doing at this point address basic bugs and minor polishings. They're really small things at this point and if anybody notices any bugs with the new changes please respond to this so that way I can fix them.

src/SSVOpenHexagon/Core/Discord.cpp
```diff
@@ -89,32 +89,83 @@ bool discord_manager::set_rich_presence_in_menu()
...
 
     discord::Activity activity{};
-    activity.SetState("In menu");
-    activity.SetDetails("...");
+    activity.SetState("Selecting Level");
+    activity.SetDetails("");
+    discord::ActivityTimestamps& currentTimestamp = activity.GetTimestamps();
+    currentTimestamp.SetStart(static_cast<int64_t>(
+        std::chrono::high_resolution_clock::now().time_since_epoch().count() /
+        1000000000));
```
SuperV1234, line 97:
```suggestion
    currentTimestamp.SetStart(static_cast<std::int64_t>(
        std::chrono::high_resolution_clock::now().time_since_epoch().count() /
        1'000'000'000));

```

Also please include `<cstdint>` here as well.

src/SSVOpenHexagon/Core/Discord.cpp
```diff
@@ -89,32 +89,83 @@ bool discord_manager::set_rich_presence_in_menu()
...
+    discord::ActivityTimestamps& currentTimestamp = activity.GetTimestamps();
+    if(!dead)
+    {
+        // Update the timestamp to show how long the current attempt is. This is
+        // shown by "MM:SS elapsed".
+        currentTimestamp.SetStart(
+            static_cast<int64_t>(std::chrono::high_resolution_clock::now()
+                                     .time_since_epoch()
+                                     .count() /
+                                 1000000000));
```
SuperV1234, line 157:
```suggestion
            static_cast<std::int64_t>(std::chrono::high_resolution_clock::now()
                                     .time_since_epoch()
                                     .count() /
                                 1'000'000'000));

```

SuperV1234 (changes requested):

SuperV1234 (approved):
LGTM. Good stuff.

PR #311, opened 2020-12-03
Favorite levels system, scrolling improvements, and bugfixes

IvoryDuke:
Reorganized the scrolling code more neatly, fixed an issue where folding huge packs would cause noticeable screen jittering.


IvoryDuke:
I realized it would be nice to have the list scroll ahead one or two levels when the selected levels are toward the top/bottom of the window.
I will take care of that tomorrow, so please don't merge the PR until then.

InfoTeddy:
> please don't merge the PR until then.

You can prevent a PR from being merged by marking it as a draft.

IvoryDuke:
> You can prevent a PR from being merged by marking it as a draft.
Thanks for the tip, I will keep it in mind.

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -332,27 +332,16 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
...
-        {
-            temp += c;
-        }
-    }
+    static const std::string tips[][2]{
+        {"HOLDING SHIFT WHILE CHANGING PACK", "SKIPS THE SWITCH ANIMATION"},
+        {"REMEMBER TO TAKE BREAKS", "OPEN HEXAGON IS AN INTENSE GAME"},
+        {"EXPERIMENT USING SWAP", "IT MAY SAVE YOUR LIFE"},
+        {"IF A LEVEL IS TOO CHALLENGING", "PRACTICE IT AT A LOWER DIFFICULTY"}
+    };
```
SuperV1234, line 340:
```suggestion
    static constexpr std::array<std::string_view, 2> tips{
        {"HOLDING SHIFT WHILE CHANGING PACK", "SKIPS THE SWITCH ANIMATION"},
        {"REMEMBER TO TAKE BREAKS", "OPEN HEXAGON IS AN INTENSE GAME"},
        {"EXPERIMENT USING SWAP", "IT MAY SAVE YOUR LIFE"},
        {"IF A LEVEL IS TOO CHALLENGING", "PRACTICE IT AT A LOWER DIFFICULTY"}
    };
```

src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -332,27 +332,16 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
...
+    const int idx = ssvu::getRndI(0, sizeof(tips) / sizeof(tips[0]));
```
SuperV1234, line 341:
```suggestion
    const int idx = ssvu::getRndI(0, tips.size());
```

include/SSVOpenHexagon/Core/MenuGame.hpp
```diff
@@ -273,7 +273,7 @@ class MenuGame
 
     // Load menu
     const hg::HGAssets::LoadInfo& loadInfo;
-    std::vector<std::string> randomTip;
+    std::string randomTip[2];
```
SuperV1234, line 276:
```suggestion
    std::array<std::string_view, 2> randomTip;
```

Please also include `<array>` and `<string_view>`. 

`std::array` is the same as a C-style array, but it has a nice container interface (e.g. `.size()`). `std::string_view` is a non-owning view over an existing range of characters which is stable in memory. Since our tips are `static`, we can use it instead.

SuperV1234 (changes requested):
LGTM, some minor C++ improvements

include/SSVOpenHexagon/Core/MenuGame.hpp
IvoryDuke: All taken care of.

IvoryDuke:
Favorite system is perhaps a bit rough as of right now but is functional.
F7 to add/remove a level to the favorites and F8 to show the favorites menu.
Favorites are saved to file when game closes.

On a side note: I did an edit on wasFocusHeld and focusHeld cause I was a bit confused on the role of the first.
If you could explain me what it was I would be happy to revert the change.

IvoryDuke:
Following a few reports of users skipping two levels using the keyboard keys I realised it occured because both the hardcoded and user defined action were activated.
So now hardcoded keys cannot be assigned a second time to an action they are alread hardcoded to.

Capture

SuperV1234 (approved):
lgtm

PR #314, opened 2021-01-03
Moved initLua before newGame() function

IvoryDuke:
Recently I noticed how sometimes the game stutters when changing selected level in the menu.
I suspect the culprit is the function initLua() which does a pretty significant amount of work and is called every time the index of the level selection is changed.
I have repositioned the lua initialization to just before a new level is initialized.

setIndex() now has a new function that goes through the lua script file to find the only two pieces of info necessary to properly draw the menu background: l_setSides and l_setRotationSpeed.
Upon applying this solution I do not notice any more stuttering in the level changes.
The function is just a file reader that looks for the onInit function and reads the two required lines. If there is a better way to handle it please let me know.

P.S: I have also renamed a variable with a misleading name and re-added the "DIFFICULTY: " string that had been removed by accident (at least I think, sorry if that was intentional and I could not figure out the reason).

### DIFF
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -97,11 +97,11 @@ private:
 
     void initAssets();
     void initInput();
-    void initLua(Lua::LuaContext& mLua);
+    void initLua();
     void initMenus();
     void playLocally();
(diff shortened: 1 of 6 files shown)


IvoryDuke:
Closed because found a much better way.

PR #315, opened 2021-01-06
Only call initLua() at boot

IvoryDuke:
Recently I noticed how sometimes the game stutters when changing selected level in the menu.
I suspect the culprit is the function initLua() which does a pretty significant amount of work and is called every time the index of the level selection is changed.
I have slightly reworked it to only be called on menu initialization. There seem to be no drawbacks from this new approach.

P.S: I have also renamed a variable with a misleading name and re-added the "DIFFICULTY: " string that had been removed by accident (at least I think, sorry if that was intentional and I could not figure out the reason).

### DIFF
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -97,7 +97,8 @@ private:
 
     void initAssets();
     void initInput();
-    void initLua(Lua::LuaContext& mLua);
+    Lua::LuaContext lua;
+    void initLua();
     void initMenus();
     void playLocally();
     std::pair<const unsigned int, const unsigned int>
--- a/src/SSVOpenHexagon/Core/MenuGame.cpp
+++ b/src/SSVOpenHexagon/Core/MenuGame.cpp
@@ -341,6 +341,7 @@ MenuGame::MenuGame(Steam::steam_manager& mSteamManager,
 
     initMenus();
     initInput();
+    initLua();
 
     //--------------------------------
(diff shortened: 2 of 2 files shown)


SuperV1234:
Seems reasonable. I can't recall any reason why this would have to be done when changing index. Any idea, @Morxemplum?

Morxemplum:
Hmmmm. I mean initLua was called every time that a level started in the original code. I think doing it every time the level changes selection is even more redundant. If initLua is supposed to define all the Lua functions the game uses, why are we even doing it multiple times? Should only have to be done once. 

Writing all the Lua functions and setting up the environment at boot seems a lot more reasonable than doing it at the start of every attempt.

Morxemplum:
I have no involvement in how Lua boots. This was always the implementation even before we began touching the source code. Sounds like an optimization that's desperately needed.

Morxemplum:
To end it off. I'm surprised that no other wrap up changes are being done to ensure a full 2.0.3 release. Unless there are some serious bugs (The only reproducible one I've seen is the double scrolling bug), I don't see why we aren't releasing 2.0.3 and updating the Lua documentation.

IvoryDuke:
> Unless there are some serious bugs (The only reproducible one I've seen is the double scrolling bug)
All menu related bugs that have been mentioned to me have been solved. This specific one has been addressed in this commit https://github.com/SuperV1234/SSVOpenHexagon/commit/76c72622787cf556966d18c739ad7938c08acc4b

SuperV1234 (approved):
lgtm

PR #316, opened 2021-01-08
3D rendering mode

IvoryDuke:
Added a second 3D mode that more convincingly gives the illusion of a 3D environment.
To activate it 3DEnabled needs to be set to 2.

Video demo: https://youtu.be/9H0KNBS1h4U

TODO:
- some levels were clearly meant to be played with the projections style of 3D and look rather ugly in this new mode, so perhaps a way do enforce a certain drawing mode needs to be added;
- whatever is found to be necessary.


SuperV1234:
Very interesting! Have you considered making this a per-level setting rather than a global one? In other words:

- We would still have a boolean global setting to enable or disable 3D effects altogether
- Each level would choose whether to be 2D, 3D (old), and 3D (new)

I think that might be a better choice so that level creators can have the most freedom

IvoryDuke:
I have actually thought about that, especially because some levels look objectively bad in the "new" 3D mode.
I have to iron out one last detail of the drawing process and then I will implement the level variable.

SuperV1234:
Cool. I am in the process of refactoring and testing #318 BTW, will merge soon

IvoryDuke:
I added a new variable to LevelData to allow level makers to select the 3D mode.
From my testing everything seems to work.

IvoryDuke:
Put this back into draft cause I realized there is a thing that needs to be addressed.

PR #318, opened 2021-01-17
Improved collision code

IvoryDuke:
Following further comments on unfair deaths I looked into the collision code of standard walls and have done some rework to provide increased reliability.
The new code passes all collision tests, presents no anomalies on curving walls, and player who have tested it say it does work better.

Process:
- as walls are updated they are tested for collision, if no collision occurs the algorithm is terminated early;
- if there is collision with a wall player can be either be pushed out of safety or killed;
- if it is pushed to safety collision with walls that were previously updated is double checked, then all remaining walls are updated and checked for collision. During this second round of collision checks player cannot be pushed again.


SuperV1234:
Extremely promising! Looking forward to review and test this in detail ASAP.

SuperV1234 (changes requested):

SuperV1234 (approved):
lgtm

SuperV1234:
Logic seems good. I tested around a bit, seems to work properly. Let's merge! Thanks!

PR #319, opened 2021-01-30
Hardcoded keys reference message

IvoryDuke:
Added a little menu option to display a message showing all the hardcoded controls.
I originally wanted to show the controls somewhere in the menu themselves but could not find a suitable spot where to place them without causing overlaps or making the screen too busy.
While a bit old school I think this is a simple and elegant solution.

Also, since F1 to F4 are no longer used I have reassigned the currently in use FX keys as shown in the picture.

capture


SuperV1234 (approved):
LGTM. This is a good improvement but I still think we need some sort of visual reminder on the level selection menu.

PR #320, opened 2021-01-30
Favorite levels visual cue

IvoryDuke:
Also improved the quick switch list scrolling (now it shows one label ahead/behind depending on the height of the labels).
NOTE: the tip can never be covered by a level label thanks to the look-ahead level scrolling system.

Capture


SuperV1234 (approved):
LGTM, this works :)

PR #321, opened 2021-01-30
Fixed pack description in favorites

IvoryDuke:
Changing level in the favorite levels submenu was not updating the pack id, and therefore the level description always defaulted to the pack at index 0.


SuperV1234 (approved):
LGTM

PR #322, opened 2021-01-31
Menu and config cleanup

IvoryDuke:
Did a private review of the code I have written for menu and config and noticed a couple things that could be rewritten in a better way, removing code duplication and implemented in a way that requires less tweaking around in case it needs to be changed.
Nothing that can be noticed from a user perspective but it makes things cleaner overall.


SuperV1234 (approved):
LGTM, making some minor changes myself after merge.

PR #323, opened 2021-02-02
Improved customwalls/regular walls collision and player push

IvoryDuke:
Reworked CPlayer::push() for customwalls to properly push player out of danger if that is possible.
Tests on levels making use of customwalls show expected results.
By default all customwalls are an instakill but they can be set to behave like regular walls and even be more lenient with two new flags.
cw_setDeadly true/false. If false they behave similarly to regular walls.
cw_setForgiving true/false. If true they are more lenient toward saving player. This is a required settings if level maker want player to slide along particularly exotic walls.

Suggested levels for testing:
- Experimental "Arcadia" (custom walls);
- Experimental "Curvetest" (curving walls);
- Experimental "Collision Test" (pulsing levels);
- Polygonal Cubics "Oh No!" (extreme customwalls);
- Polygonal Cubics "How?!?" (extreme customwalls);
- Polygonal Cubics "Misdirection" (customwalls on level with really high amount of sides);
- Attached level zip (standard walls on level with really high amount of sides).

[sandbox.zip](https://github.com/SuperV1234/SSVOpenHexagon/files/5962613/sandbox.zip)


IvoryDuke:
I ran into an unexpected issue which will require a little more time to figure out.
I am very close to a viable solution but need to address a rather annoying bug before I update the PR.

_RELEASE/Assets/favoriteLevels.json
```diff
@@ -2,10 +2,7 @@
 {
 	"ids" : 
 	[
-		"LoFiOHv1_Jocii_Chill_Open_Hexagon_to_Calm_the_Vibe_1_LoFi01",
-		"ohvrvanilla_vittorio_romeo_cube_1_apeirogon",
-		"ohvrvanilla_vittorio_romeo_cube_1_commando",
-		"ohvrvanilla_vittorio_romeo_experimental_1_curvetest",
-		"ohvrvanilla_vittorio_romeo_experimental_1_curvetest2"
+		"ohvrvanilla_vittorio_romeo_experimental_1_arcadia",
```
SuperV1234, line 5: Favorite levels should be saved on a per-profile basis, but I'm fine with that being done for a future update. Also this file should be cleared up before merging.

include/SSVOpenHexagon/Components/CPlayer.hpp
```diff
@@ -21,9 +23,13 @@ class CCustomWall;
...
 {
 private:
-    sf::Vector2f startPos;
-    sf::Vector2f pos;
-    sf::Vector2f lastPos;
+    sf::Vector2f startPos;      // Position at start of the level.
+    sf::Vector2f pos;           // Actual position of player.
+    sf::Vector2f prePushPos;    // Position before player is pushed by a wall.
+                                // Unline pos it is not updated after a successful
+                                // wall push.
```
SuperV1234, line 30:
```suggestion
    sf::Vector2f prePushPos;    // Position before the player is pushed by a wall.
                                // Unlike `pos` it is not updated after a successful
                                // wall push.
```

include/SSVOpenHexagon/Components/CPlayer.hpp
```diff
@@ -21,9 +23,13 @@ class CCustomWall;
...
+    sf::Vector2f lastPos;       // Position of player in the previous frame, adjusted
+                                // according to the current frame's radius.
```
SuperV1234, line 32:
```suggestion
    sf::Vector2f lastPos;       // Position of the player in the previous frame, adjusted
                                // according to the current frame's radius.
```

include/SSVOpenHexagon/Components/CPlayer.hpp
```diff
@@ -34,6 +40,11 @@ class CPlayer
 
     bool dead;
     bool justSwapped;
+    bool forcedMove;
+
+    float radius;
+    float maxSafeDistance;
+    float currentSpeed;
```
SuperV1234, line 47: These fields should be documented as well, otherwise it will be hard to keep track of everything a few weeks from now :)

include/SSVOpenHexagon/Components/CWall.hpp
```diff
@@ -28,31 +28,32 @@ class CWall
     float hueMod;
     bool killed;
 
-public:
-    CWall(HexagonGame& mHexagonGame, const sf::Vector2f& mCenterPos, int mSide,
-        float mThickness, float mDistance, const SpeedData& mSpeed,
-        const SpeedData& mCurve);
-
-    void update(const HexagonGame& mHexagonGame, const ssvu::FT mFT);
+    bool outOfPlayerRadius;
```
SuperV1234, line 31: Ditto, please document

include/SSVOpenHexagon/Components/CWall.hpp
```diff
@@ -67,12 +68,40 @@ class CWall
...
+        // If the wall is too far from the center is cannot cause collision.
+        // If it is and both distance and pointInPolygon() calculations are executed
+        // we lose performance, If not (which is the case in the vast majority of
+        // any level's runtime) we only do a much faster algorithm.
+        if(ssvs::getMagSquared((vertexPositions[0] + vertexPositions[1]) / 2.f -
+            mCenterPos) > mRadiusSquared)
+        {
+            outOfPlayerRadius = true;
+            return false;
+        }
```
SuperV1234, line 84: This seems like a nice little optimization. Have you actually measured the impact? Would be interesting to know. 
Also wondering if we can avoid the call to `getMagSquared` somehow and avoid the extra multiplication, maybe we could cache that value once per frame?

include/SSVOpenHexagon/Utils/PointInPolygon.hpp
```diff
@@ -12,12 +12,12 @@ template <typename TC, typename T>
     const TC& mVertices, T x, T y) noexcept
 {
     bool result{false};
+    auto size{mVertices.size()};
```
SuperV1234, line 15:
```suggestion
    const auto size{mVertices.size()};
```

include/SSVOpenHexagon/Utils/PointInPolygon.hpp
```diff
@@ -29,4 +29,26 @@ template <typename TC, typename T>
     return result;
 }
 
+template <typename TC, typename T>
+[[gnu::always_inline, gnu::pure, nodiscard]] inline bool pointInPolygonPointers(
+    const TC& mVertices, T x, T y) noexcept
+{
+    bool result{false};
+    auto size{mVertices.size()};
```
SuperV1234, line 37:
```suggestion
    const auto size{mVertices.size()};
```

include/SSVOpenHexagon/Utils/Utils.hpp
```diff
@@ -43,19 +43,39 @@ inline void uppercasify(std::string& s)
     return s;
 }
 
-inline float getFontHeight(sf::Text& font)
+[[nodiscard, gnu::pure]] inline float getFontHeight(sf::Text& font)
```
SuperV1234, line 46:
```suggestion
[[nodiscard]] inline float getFontHeight(sf::Text& font)
```

`gnu::pure` can be a bit problematic with references involved. I am not sure it can be applied correctly here, so please remove it.

include/SSVOpenHexagon/Utils/Utils.hpp
```diff
@@ -43,19 +43,39 @@ inline void uppercasify(std::string& s)
...
 {
     font.setString("A");
     return ssvs::getGlobalHeight(font);
 }
 
-inline float getFontHeight(sf::Text& font, const unsigned int charSize)
+[[nodiscard, gnu::pure]] inline float getFontHeight(sf::Text& font, const unsigned int charSize)
```
SuperV1234, line 52:
```suggestion
[[nodiscard]] inline float getFontHeight(sf::Text& font, const unsigned int charSize)
```

Ditto

include/SSVOpenHexagon/Utils/Utils.hpp
```diff
@@ -43,19 +43,39 @@ inline void uppercasify(std::string& s)
...
+    const sf::Vector2f& p2, const float mRadiusSquared);
+
+bool getLineCircleClosestIntersection(
+    sf::Vector2f& mIntersection, const sf::Vector2f& mPos,
+    const sf::Vector2f& p1, const sf::Vector2f& p2,
+    const float mRadiusSquared);
+
+[[nodiscard, gnu::pure]] inline float fastSqrt(const float &n) 
+{
+    static union{int i; float f;} u;
```
SuperV1234, line 74:
```suggestion
    union{int i; float f;} u;
```

Why `static`?

include/SSVOpenHexagon/Utils/Utils.hpp
```diff
@@ -43,19 +43,39 @@ inline void uppercasify(std::string& s)
...
+unsigned int getLineCircleIntersection(sf::Vector2f& i1,
+    sf::Vector2f& i2, const sf::Vector2f& p1,
+    const sf::Vector2f& p2, const float mRadiusSquared);
+
+bool getLineCircleClosestIntersection(
+    sf::Vector2f& mIntersection, const sf::Vector2f& mPos,
+    const sf::Vector2f& p1, const sf::Vector2f& p2,
+    const float mRadiusSquared);
+
+[[nodiscard, gnu::pure]] inline float fastSqrt(const float &n)
```
SuperV1234, line 72: Did you measure that this is actually faster than `std::sqrt`?

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -159,112 +152,242 @@ void CPlayer::kill(HexagonGame& mHexagonGame)
...
-    if(!getJustSwapped())
+template <typename Wall>
+[[nodiscard]] bool CPlayer::checkWallCollisionEscape(const Wall& wall,
+    sf::Vector2f& mPos, const float mRadiusSquared)
+{
+    // To find the closest wall side we intersect the circumference of the possible
+    // player positions with the sides of the wall. We use the intersection closest
+    // to the player's position as post collision target.
+    // If an escape route could not be found player is killed.
+
```
SuperV1234, line 167: This is all pretty arcane for me but I'm just gonna trust you :)

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -159,112 +152,242 @@ void CPlayer::kill(HexagonGame& mHexagonGame)
...
-    const sf::Vector2f testPos{
-        ssvs::getOrbitRad(startPos, testAngle, mHexagonGame.getRadius())};
-
-    // If there is overlap even after compensation kill without updating
-    // position, as there is no benefit in doing it.
-    if(wall.isOverlapping(testPos))
+    // ...otherwise make testPos the position of the previous frame plus
+    // the curving wall's velocity, and check an escape on that position.
+    // Using the previous frame's position is essential for levels with
+    // a really high amount of sides.
```
SuperV1234, line 247: Why? Would be nice to know in the comment

src/SSVOpenHexagon/Components/CPlayer.cpp
```diff
@@ -159,112 +152,242 @@ void CPlayer::kill(HexagonGame& mHexagonGame)
...
+    if(wall.isOverlapping(testPos) ||
+       !checkWallCollisionEscape(wall, testPos, mRadiusSquared))
+    {
+        return true;
+    }
+
+    // If player survived apply test position and add a little padding.
+    pos = testPos + ssvs::getNormalized(testPos - prePushPos) * collisionPadding;
+    angle = ssvs::getRad(pos);
+    updatePosition(mHexagonGame, mFT);
```
SuperV1234, line 390: Ditto, this is also pretty arcane to me (the maths), but I'm gonna trust (and test)

src/SSVOpenHexagon/Components/CWall.cpp
```diff
@@ -79,17 +85,15 @@ void CWall::moveTowardsCenter(HexagonGame& mHexagonGame,
         ssvs::moveTowards(vp, mCenterPos, speed.speed * 5.f * mFT);
     }
 
-    if(pointsOnCenter > 3 || pointsOutOfBounds > 3)
+    if(pointsOnCenter == 4 || pointsOutOfBounds == 4)
```
SuperV1234, line 88: Why was this changed? Maybe there could be a weird scenario where a weirdly shaped wall has its vertices counted multiple times... or maybe it can never happen...?

SuperV1234 (changes requested):
Looks generally good! Will have to test a lot and make sure everything works, but this is awesome. Just a few things to improve/fix

include/SSVOpenHexagon/Utils/Utils.hpp
IvoryDuke: It is a well known algorithm that has been used since the Quake engine.
I have not measured myself how it compares to std::sqrt, I will inquire when possible.

src/SSVOpenHexagon/Components/CWall.cpp
IvoryDuke: I made it == 4 cause I always found it to be somewhat misleading to read > 3, which to me implied there could be polygons with an arbitrary number of vertices higher than 3.
I do not mind reverting it though.

include/SSVOpenHexagon/Components/CWall.hpp
IvoryDuke: I am not sure how we could avoid using getMagSquared(), I am already skipping the unnecessary step of running sqrt on it since if x > y => x * x > y * y but that is as far as I think we can go without running in inaccurate calculations.
Caching would provide no benefits since that specific overload is called only the first collision check of a frame, the second one relies on
```
    [[gnu::always_inline, nodiscard]] bool isOverlapping(
        const sf::Vector2f& mPoint) const noexcept
    {
        if(outOfPlayerRadius)
        {
            return false;
        }

        return Utils::pointInPolygon(vertexPositions, mPoint.x, mPoint.y);
    }
```
I will see to measure the impact one of the next days.

include/SSVOpenHexagon/Utils/Utils.hpp
IvoryDuke: I copied-pasted the algorithm from a game development website. Gonna make it static.

include/SSVOpenHexagon/Utils/Utils.hpp
IvoryDuke: Run a simple test, this example code returns 0 for the duration of fastSqrt and 1 for std::sqrt
```
#include <iostream>
#include <chrono>
#include <cmath>

inline float fastSqrt(const float &n) 
{
    union{int i; float f;} u;
    u.i = 0x5F375A86 - (*(int*)&n >> 1);
    return (int(3) - n * u.f * u.f) * n * u.f * 0.5f;
}

int main()
{
    const float num = 38725673265436589345.324677;
    
    auto t1 = std::chrono::high_resolution_clock::now();
    const auto f = fastSqrt(num);
    auto t2 = std::chrono::high_resolution_clock::now();
    std::cout << std::chrono::duration_cast<std::chrono::microseconds>( t2 - t1 ).count() << std::endl;
    
    auto t3 = std::chrono::high_resolution_clock::now();
    const auto g = std::sqrt(num);
    auto t4 = std::chrono::high_resolution_clock::now();
    std::cout << std::chrono::duration_cast<std::chrono::microseconds>( t4 - t3 ).count() << std::endl;
    
    return 0;
}
```

IvoryDuke:
I have added explanatory comments on all requested parts of the code and have applied suggested tweaks.

include/SSVOpenHexagon/Components/CWall.hpp
IvoryDuke: Example code returns 2 0
` ` ` 
#include <iostream>
#include <chrono>
#include <array>

struct Point
{
    float x;
    float y;
};

inline bool pointInPolygon(const std::array<Point, 4>& points, float x, float y) noexcept
{
    bool result{false};

    for(unsigned int i{0}, j{3}; i < 4; j = i++)
    {
        const auto& vI{points[i]};
        const auto& vJ{points[j]};

        if(((vI.y > y) != (vJ.y > y)) &&
            (x < (vJ.x - vI.x) * (y - vI.y) / (vJ.y - vI.y) + vI.x))
        {
            result = !result;
        }
    }

    return result;
}

inline float magSquared(Point& p) noexcept
{
    return p.x * p.x + p.y * p.y;
}

int main()
{
    auto t1 = std::chrono::high_resolution_clock::now();
    std::array<Point, 4> points = {Point{0.f, 0.f}, Point{5.f, 0.f}, Point{10.f, 22.f}, Point{-2.f, 6.f}};
    const auto b = pointInPolygon(points, 4.f, 6.f);
    auto t2 = std::chrono::high_resolution_clock::now();
    std::cout << std::chrono::duration_cast<std::chrono::microseconds>( t2 - t1 ).count() << std::endl;

    auto t3 = std::chrono::high_resolution_clock::now();
    Point p = {20.f, 16.f};
    const auto f = magSquared(p);
    auto t4 = std::chrono::high_resolution_clock::now();
    std::cout << std::chrono::duration_cast<std::chrono::microseconds>( t4 - t3 ).count() << std::endl;

    return 0;
}
` ` `

IvoryDuke:
Replaced forgiving status with the possibility for users to choose which side should definitively kill player.
This both solves a few rare unjustified deaths and allows level makers to make more precise and deliberate choices on the behavior of the walls.
The reason why collision on standard walls is precise and reliable is that we know for sure that the only side of it that should kill the player is the one closer to the center (the one delimited by vertexes 0 and 1). This puts more work on level makers but uniforms behavior between custom and regular walls when used correctly.

SuperV1234:
> Replaced forgiving status with the possibility for users to choose which side should definitively kill player.

Do you think you can finalize the work on custom walls soon? Or maybe we should release it as experimental? It's been a while since a new version release :)

IvoryDuke:
I actually consider this to be final, the thing I just did randomly popped into my mind today.
It was very functional even before, but this makes it even more reliable.

SuperV1234 (approved):

PR #324, opened 2021-02-05
fix collision on levels with a really high amount of sides

IvoryDuke:
The fix: in CPlayer::push() I was reusing lastAngle and lastPos to store the position to be tested for overlap.
Problem is, to then figure put where the player should have been repositioned we need to use the actual position in the previous frame, not the test position i stored in the variables a few lines before.
This was an issue on levels with an outrageous amount of sides (more than 50) where pinpoint accuracy was essential.

Instead of reusing lastPos I added a dedicated variable in the function (testPos) to store the position to be tested after the movement compensation.


SuperV1234 (approved):
LGTM

PR #325, opened 2021-02-24
Fix favorites menu scroll

IvoryDuke:
The favorites menu scroll looked janky when there were more level labels than they could fit on screen.
This solves the issue.


SuperV1234 (approved):
LGTM

PR #326, opened 2021-02-26
Menu levelDataIds optimization

IvoryDuke:
Noticed the menu performs a rather expensive copy of std::vector when changing pack.
I have optimized it to instead point to the vectors stored in assests


include/SSVOpenHexagon/Global/Assets.hpp
```diff
@@ -71,6 +71,16 @@ class HGAssets
     [[nodiscard]] bool loadPackData(const ssvufs::Path& packPath);
     [[nodiscard]] bool loadPackInfo(const PackData& packData);
 
+    void setLevelFavoriteFlag(const std::string& mAssetId, const bool flag)
+    {
+        getEditLevelData(mAssetId).favorite = flag;
+    }
+
+    bool checkLevelIDPurity(std::string& mAssetId)
```
SuperV1234, line 79:
```suggestion
    [[nodiscard]] bool checkLevelIDPurity(const std::string& mAssetId)
```

The name is also a bit confusing...

SuperV1234 (changes requested):
I need to have a better look at this tomorrow when I'm fresh to check that everything is good in terms of memory. Also, I want to remind you that favorite levels should be on a per-profile basis...  Do you think it will be easy to get there from the current state?

SuperV1234 (approved):

PR #332, opened 2021-04-02
Cleaner input movement update

IvoryDuke:
A cleaner (I think) way to process the user inputs for the direction.
It likely does not provide better performance than the current implementation but I personally prefer it over the chain of if/else if statements that are currently in place.


SuperV1234:
Thanks! Already had made some changes for input processing in the v2.0.5 branch, will apply some of your improvements manually.

PR #334, opened 2021-04-03
Fix focus resetting background animation in main menu

IvoryDuke:
As per title, now focus presses only effect the menu when in Level Selection.


SuperV1234:
LGTM

PR #337, opened 2021-04-11
Multithreaded assets load

IvoryDuke:
Split the boot assets load among a customizable amount of threads for much faster load times.
All data vectors have been secured behind mutexes and the load is articulated in two phases to avoid race conditions and inconsistencies.


SuperV1234 (approved):
LGTM, thanks! Will make some final changes myself like using `std::lock_guard` ;)

PR #340, opened 2021-05-23
Changes towards OSX building

duck-37:
1. Make CMake work with OSX
2. Add needed third-party libraries

TODO: Will not compile because of problems with std::chrono.


SuperV1234:
Hey @duck-37, I would appreciate OSX support! Are you still working on this?

duck-37:
> 
> 
> Hey @duck-37, I would appreciate OSX support! Are you still working on this?

Not at the moment as I don't know exactly where to continue here. The only remaining issue is the conflict between high_resolution_clock and system_clock, and that means I need to modify the database code to use system_clock. Though it's mostly abstracted away, I don't know whether it'll cause issues somewhere else down the line.

PR #341, opened 2021-05-23
Add bash script to copy build artifacts

Morxemplum:
This is a bash script that was designed to copy the build artifacts and supply all the necessary DLLs to _RELEASE for the game to run properly, along with the executables themselves.

The bash script asks for the user's `msys64` folder, as the msys system has some DLLs that are needed to be copied.

### DIFF
--- a/build/copyArtifacts.bash
+++ b/build/copyArtifacts.bash
@@ -1,0 +1,54 @@
+#!/bin/bash -u
+
+# We need the user to specify an argument for where their MSYS folder is
+if (( $# >= 1 )); then
+	msysLoc=$1
+else
+	echo "Please enter the path to your installed MSYS folder:"
+	read msysLoc
+fi
+
+if [[ $msysLoc == "" ]]; then
+	echo "ERROR: No path is provided. Exiting" 1>&2
+	exit 1
+fi
+
+if [[ ! -d ${msysLoc} ]]; then
+	echo "ERROR: The path provided doesn't exist" 1>&2
+	exit 1
+fi
+
+# Locate the _deps and RELEASE folders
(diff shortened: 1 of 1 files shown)


SuperV1234:
Looks good. I just realized I had something similar in the `build/` folder:
https://github.com/SuperV1234/SSVOpenHexagon/blob/master/build/copylibs.sh

Is there anything you want to take from there? Or maybe get rid of it and avoid duplication? Note that `copylibs.sh` is used by:
- https://github.com/SuperV1234/SSVOpenHexagon/blob/master/build/make_debug_client_win10_msys_2_copy.sh
- https://github.com/SuperV1234/SSVOpenHexagon/blob/master/buildrel/make_release_client_win10_msys_2_copy.sh
- https://github.com/SuperV1234/SSVOpenHexagon/blob/master/buildlx/make_release_server_vbox_2_copy.sh

Which are scripts I personally use for building release and debug version of OH and its server, which I want to keep. I am also fine to merge the PR as is, just let me know.

Morxemplum:
The key difference in mine is that it asks the user for the msys64 folder because we don't know where it might be. The one thing that interests me is that your script copies it to a test folder, and I wonder what the purpose of the test folder is.

SuperV1234:
> I wonder what the purpose of the test folder is.

Running automated test drivers as part of the usual compile&run process during development.

Morxemplum:
So it's responsible for CI? Gotcha
I don't know how that will go with my Bash script because it requires user input and also has an "Enter to Exit" at the end of the script.

SuperV1234:
I will merge this as is, we can cleanup the whole build process later. The important thing is that you're able to build

PR #344, opened 2021-05-25
Add Wall Color Override to style JSONs

Morxemplum:
Another little pull request that I want to add to the game. This one adds a wall color override, which I should've done back in 2.0.3. I've even made a nice little experimental level with a style that I created using the wall color override, and it shows how you can make some nice stuff. I think it's nice you can have some wall color customization without the need to use custom walls.

I'm coming back to source code development with small little pull requests like these before I have to start working on my bigger changes.


SuperV1234 (approved):
lgtm, nice!

PR #345, opened 2021-05-25
Fix Black and White showing color in level / menu

Morxemplum:
I know that a lot of people don't pay attention to Black And White mode, but I wanted to enable it so I could have some pictures for some artwork, and noticed how Black And White hasn't adjusted to fit some of the new style changes that have been introduced. This pull request fixes those bugs and also tries its best to provide a convincing Black And White look to the level selection menu as well.

Still, a bug fix is a bug fix.


Morxemplum:
In addition, it gives me another glimpse on how the menu is written, and the way the menu does colors is so messy...

Morxemplum:
Keep in mind it's still not true BnW, as swap color and death effects still exhibit color.

SuperV1234 (approved):
LGTM

PR #346, opened 2021-06-02
Add Music Timeline (Global Timeline) to Open Hexagon to enhance rhythm aspects

### LINKED ISSUE #335: Make a Timeline that is controlled specifically by a level's music
I've been brainstorming on how to do this for the past few hours. And I'm raising this as an issue so that I can remind myself to do it when I got the time to do so (which won't be for a bit).

I've noticed that one Steam Review, in particular, mentions the game's lack of rhythm, and how it degrades the quality of the game. To be fair, I see where he's coming from. Super Hexagon incorporates a sense of rhythm by syncing the beat pulse/zoom to the music itself. When you look at other arcade games that could attract the same audience to Open Hexagon, they all have a sense of rhythm to them (a lot of them being classified as rhythm games). Open Hexagon isn't necessarily a rhythm game when we take a look at the default packs, because that's not the intended purpose of the default packs. However, there can be some additional elements of rhythm incorporated into the default levels by having levels mimic Super Hexagon's mechanic.  But most importantly, pack developers can create levels that can make Open Hexagon fit the definition of a rhythm game.

However, when it comes to thinking of a solution, Super Hexagon's rhythm mechanics are only an edge case. I've also been noticing that a lot of Open Hexagon pack developers are incorporating rhythm into their levels, most notably by using timed events on the event timeline. However, using the event timeline to incorporate rhythm has restrictions:

- Music files can only have one starting position, as a completely different starting position can invalidate an entire sequence of events completely.
- They do not take into account that music can loop. Sure, you could definitely shove all of the events into a function and call it every time the song loops, but that means the durations are going to be relative instead of absolute, and that can be more tedious to calculate. 
- Music-related events can be mixed up with actual events if the pack developer doesn't properly organize their code, which can lead to a bloated and unorganized event timeline.

We have been making many measures for Open Hexagon pack developers to help their level incorporate rhythm and sync to their music, but I believe this is a feature that helps change Open Hexagon for those who want to add rhythm more than any changes we've made so far. This will make their job significantly easier, and can possibly be paradigm-shifting in terms of what levels can be on the Workshop.

### The Music Timeline
I thought the best way to incorporate rhythm into a level is for there to be a completely new timeline, apart from the pattern timeline and the event timeline we know and love. I'm calling this timeline the music timeline, for self-explanatory reasons. Here are some unique features that this Music Timeline will have.

- Like the pattern and event timelines, it can evaluate blocks of Lua code
- The "cursor" will be on where the music is currently playing. The cursor for the event timeline, for reference, is based on the time elapsed for an attempt. However, where the song is playing, the cursor will be there. So if a song starts 70 seconds in when a level starts, the cursor will be at 70 seconds instead of 0.
- This will be an **absolute** timeline. The pattern and event timelines are relative timelines, as they are based on when the level starts. However, music must be absolute. Since the starting position of a song can be randomized, we cannot rely on a relative timeline because it will take too many calculations to adjust for new starting positions. 
    - Since this is an absolute timeline, waiting functions are not needed. If we want relativity, we would use Lua arithmetic to apply relative transformations to an absolute value. 
- Unlike the other timelines, this timeline will not remove events after using them. So when a song loops, a string of events can be re-executed to ensure consistency. Or, if you decide to rewind or change the timestamp of the music, it will replay the events alongside the music. 

### The Lua implementation
First, we need to establish which prefix the music timeline functions will be using. Since this is audio-related, `a_` is the best choice here. Now, what functions will be added?

- `a_evalTo(float mSeconds, string mLua)`: Similar to `t_eval` and `e_eval`, this function will add a Lua block of code onto the music timeline at `mSeconds`. Since the music timeline is absolute, it needs an additional argument to specify where in the timeline it must be placed. 
- `a_clear()`: Clears all events on the music timeline.

Seriously, that's it. Again, the waiting functions are not needed since we're working with an absolute timeline. Perhaps later down the road, we can add functions relating to getting all of the code blocks that will be executed at a specific timestamp. But that is for another time.

I'm looking forward to implementing this, as this will definitely be a game-changer for not only pack developers but the players as well, as it can improve their experience towards the game. Again, I can't start on this for some time because I'm occupied, but when I can start working on the source code, I'll be glad to work on this feature. :)

Morxemplum:
Closes #335.

I've managed to implement the music timeline. Turns out that the timeline class that the pattern, event, and custom timelines use are based on `timeline2`, which is built to function as a relative timeline. So I had to actually create a brand new class that represents a global timeline. This was the trickiest part because there are things that the global timeline class does differently from timeline2. The global timeline:
- Uses integers for timepoints instead of Chrono timepoints for simplicity.
- Doesn't have any waiting functions, because waiting functions don't serve a purpose in a global timeline.
- Uses a map to keep track of actions instead of a vector. The map contains vector values to allow for multiple events at the same time.
- Can account for any abrupt changes in the time by clearing the previous timepoint
- Doesn't remove events after executing them, allowing for repeated execution.

The global timeline is used to create the music timeline, and it functions almost like your typical timeline. The main difference comes in when you have to time your events. Instead of a wait function, you have to specify a specific time when the event will happen. If you want to use relative transformations, you can use simple math to recreate a waiting function. 

There is now a "music pointer", which relays the current time that the song is playing. The music pointer will follow along with the music pitch and music changes. The music pointer works off milliseconds, rather than frame time (to account for precision in music). Due to an OpenAL limitation, the music pointer updates every 20ms. If precision is really important, I will go back and try to redo this pointer to work off the game logic instead of the SFML logic. 

Speaking of Lua functions, here are all the new Lua functions that this pull request adds:
- `a_getMusicSeconds`: Grabs the value of the music pointer in seconds
- `a_getMusicMilliseconds`: Grabs the value of the music pointer in milliseconds. Unlike the previous function, this returns an integer in case you're worried about floating-point error.
- `a_evalTo`: Puts a Lua block of code onto the music timeline at a specific timestamp in seconds.
- `a_evalToMs`: Puts a Lua block of code onto the music timeline at a specific timestamp in milliseconds.
- `a_clearTimepoint`: Removes all events from that specific time point in the music timeline.
- `a_clearTimepointMs`: Millisecond variant.
- `a_clear`: Removes all events from the music timeline.

There were a few more that I added than originally intended. I wanted to add clearing specifically for a timepoint for quality of life, and I added millisecond variants to ensure precision.

### DIFF
--- a/include/SSVOpenHexagon/Core/HGStatus.hpp
+++ b/include/SSVOpenHexagon/Core/HGStatus.hpp
@@ -31,6 +31,7 @@ private:
     double currentPause{0.1 * 60};       // Current pause time
     double currentIncrementTime{};       // Time since last increment
     float customScore{};                 // Value for alternative scoring
+    int32_t musicPointer{};              // Current time in level music
 
 public:
     float pulse{75};
@@ -83,6 +84,15 @@ public:
 
     // Accumulate the time spent in a frame into the total
     void accumulateFrametime(const double ft) noexcept;
+
+    // Get the current music time in milliseconds
+    [[nodiscard]] int32_t getMusicTime() noexcept;
+
+    // Get the current music time in seconds
+    [[nodiscard]] double getMusicTimeSeconds() noexcept;
+
+    // Update the music pointer with the new time
+    void updateMusicTime(int32_t newTime) noexcept;
 
     // Update the custom score
     void updateCustomScore(const float score) noexcept;
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -15,6 +15,7 @@
 #include "SSVOpenHexagon/Utils/LuaWrapper.hpp"
 #include "SSVOpenHexagon/Utils/FastVertexVector.hpp"
 #include "SSVOpenHexagon/Utils/Timeline2.hpp"
+#include "SSVOpenHexagon/Utils/TimelineGlobal.hpp"
 #include "SSVOpenHexagon/Components/CCustomWallManager.hpp"
 
 #include <SSVStart/GameSystem/GameSystem.hpp>
@@ -141,6 +142,9 @@ private:
 
     Utils::timeline2 eventTimeline;
     Utils::timeline2_runner eventTimelineRunner;
+
+    Utils::timelineGlobal musicTimeline;
+    Utils::timelineGlobal_runner musicTimelineRunner;
 
     Utils::timeline2 messageTimeline;
     Utils::timeline2_runner messageTimelineRunner;
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -89,6 +89,7 @@ private:
     sf::Font& font;
 
     Audio* audio;
+    MusicData::Segment segment;
 
     const LevelData* levelData;
 
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -303,6 +303,7 @@ private:
     void updateWalls(ssvu::FT mFT);
     void updateIncrement();
     void updateEvents(ssvu::FT mFT);
+    void updateMusic(ssvu::FT mFT);
     void updateLevel(ssvu::FT mFT);
     void updateCustomWalls(ssvu::FT mFT);
     void updatePulse(ssvu::FT mFT);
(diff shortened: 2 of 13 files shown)


SuperV1234:
If I am understanding correctly, the music timeline might change behavior depending on the random music segment that's being played, for levels with multiple music segments. Is that right?

If so, we need to devise a strategy for replay synchronization first, as replays do not currently store the music segment as it hasn't been significant to gameplay until now.

Morxemplum:
You do have a fair point. We probably should make replays hold music segment information. This is also something that has been heavily requested by players as well.

Morxemplum:
I have no idea what is causing that weird segfault and is causing the GitHub Actions to fail. This happened since the first push onto the branch and I have no reason to believe that a simple music pointer and music timeline can lead to a segfault with a replay test case. The two seem completely unrelated and I wish the Actions were more verbose with the error information so I can understand what is causing it.

fentras:
Merging with latest `master` (as of writing this comment) fixes the problem.

https://user-images.githubusercontent.com/26721650/152653610-13870dda-0082-421b-82c6-5deb80c0fde0.mp4

Morxemplum:
Considering the activity has been stale and a more modern pull request has failed to merge, no point in keeping this open.

PR #348, opened 2021-07-07
Fix Steam stats on platforms without API bug

MultHub:
The Steam API appears to be bugged only on **some** platforms. On platforms not affected by this bug (including Linux and possibly others), getting and setting stats would fail. This PR works around this platform-specific issue by trying to get/set stats as floats and falling back to integers if that fails.


SuperV1234 (approved):
lgtm

SuperV1234:
Thanks!

PR #349, opened 2021-07-11
Allow binary file reading using io.open(filename, "rb")

MultHub:
On **some** platforms (including Windows), `io.open(filename, "r")` (currently the only supported mode) performs potentially unwanted newline conversion, forcing pack developers to work around this restriction using base64 or similar binary-to-text encoding formats. This PR allows pack developers to read files in binary mode using `io.open(filename, "rb")`.


Morxemplum:
I'm for it. Might help cut down on some file sizing if it is necessary.

SuperV1234:
LGTM, merging into 2.0.7 branch

PR #368, opened 2022-02-08
Copy artifacts update

PKPenguin321:
sfml-build dlls are now on 3, up from 2, so the current copyArtifacts.bash doesn't locate them correctly and will fail to copy them. This fixes that


vittorioromeo:
Thanks!

PR #369, opened 2022-02-09
Organize replays into folders

PKPenguin321:
A really small change that makes it so replays are now saved under `Replays/levelId/difficultyMult/`, rather than just `/Replays`.
Example: A replay for Incongruence at 0.5x would be saved under `Replays\ohvrvanilla_vittorio_romeo_hypercube_1_incongruence\0.5x\replayName.ohr.z`

Tested and confirmed functional, and a quick code read indicates that it should not affect uploading to leaderboards (I seem to be unable to actually test this locally however, I assume logging in to the server is disabled via the github build)


vittorioromeo (approved):
LGTM, thanks for this! Not sure why I didn't merge it earlier....

PR #374, opened 2022-07-10
Fix faulty last pack menu stretch

IvoryDuke:
Opened OH to check on the last update and noticed a bug in the level selection menu, as shown in the video.
(After the stretch, at the bottom of the screen there should be the selected level, which is the last one of the last pack)
https://youtu.be/Bz_ul7hOxa0

I haved fixed it and did some testing, everything seems to behave as it should.

### DIFF
--- a/include/SSVOpenHexagon/Core/MenuGame.hpp
+++ b/include/SSVOpenHexagon/Core/MenuGame.hpp
@@ -456,6 +456,8 @@ private:
 
     void checkWindowTopScroll(
         const float scroll, std::function<void(const float)> action);
+	bool checkWindowTopScrollWithResult(
+        const float scroll, std::function<void(const float)> action);
 
     void checkWindowBottomScroll(
         const float scroll, std::function<void(const float)> action);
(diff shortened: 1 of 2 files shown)


IvoryDuke:
I'll give a concise answer:
- at the core there is a state machine, depending on the value of the state a different menu is drawn;
- at boot (initMenus()) all menus and menu elements are allocated in memory;
- there are different drawing functions depending on the menu that needs to be drawn. At the end it's all a bunch of text elements with quads drawn behind them (quads drawing uses the same allocation and drawing methods of the in game walls). For example the main menu has a quad for each item, option menus have a single quad drawn behind all the menus;
- the xxAction() functions allow menu navigation and activate the menu items depending on the state machine value;
- the level selection menu has a right and left side. The left retrieves the info of the selected level and draws them one by one from the top to the bottom, properly formatting the level description to fit in the assigned space. The right is the full level selection, first the list of the levels of the current pack and then all the labels;
- most of the remaining code is for the menu transitions and folding/stretching effects.

Basically: get menu->go to the necessary drawing function->process any eventual transition effects->draw the quads and the text in the right order (text on tp of quads).

Syyrion:
Okay thanks. I think I can figure things out from here. (UI programming always sucks...)

vittorioromeo:
Thanks for this. I have no time to test it now, but will do ASAP.

vittorioromeo:
Thanks!

PR #375, opened 2022-07-16
Fix the shuffle function

Syyrion:
The original shuffle function was a wrong implementation of the Fisher-Yates shuffle.


vittorioromeo (approved):
LGTM, thanks

PR #376, opened 2022-07-16
add lua function to get scores

Bauumm:
This would allow for progression in level packs. I wasn't sure if you should be able to only get scores of your own pack, but since I don't think it's an issue if you can see other scores and it's easier to implement that way too, I think it's fine.


vittorioromeo:
Generally not a fan of sharing state between levels, and having to manually pass level ids around sounds a bit fragile. My idea is that each level should be as independent as possible, at least right now. If we want to implement progression, we should have some better ways of doing that -- e.g. pack-based state that is customizable and retained on a per-profile basis.

PR #378, opened 2022-09-05
Add 3D layer offset as style property

Bauumm:
Originally I made this to shift 3D layers up by one for 1.92 parity (for the level converter), but it turned out that quite a lot of fun effects can be done with this, so I also added the ability to render 3D layers above the main layer and a way to mirror the alpha falloff as well as the option to render the main layer on top of all 3D layers.


include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -387,6 +387,7 @@ class HexagonGame
     void drawTrailParticles();
     void drawSwapParticles();
     void drawImguiLuaConsole();
+    void drawMainLayer(const auto getRenderStates);
```
vittorioromeo, line 390: Unused, right?

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -273,6 +327,10 @@ void HexagonGame::draw()
     render(pivotQuads, getRenderStates(RenderStage::PivotQuads));
     render(playerTris, getRenderStates(RenderStage::PlayerTris));
 
+    render(wallQuads3DTop, getRenderStates(RenderStage::WallQuads3D));
+    render(pivotQuads3DTop, getRenderStates(RenderStage::PivotQuads3D));
+    render(playerTris3DTop, getRenderStates(RenderStage::PlayerTris3D));
```
vittorioromeo, line 332: Can this also be wrapped in `if (renderAbove) { ... }`?

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -245,11 +291,19 @@ void HexagonGame::draw()
...
+                if(renderingAbove)
+                {
+                    playerTris3DTop[k].position += newPos;
+                    playerTris3DTop[k].color = overrideColor;
+                }
+                else
+                {
+                    playerTris3D[k].position += newPos;
+                    playerTris3D[k].color = overrideColor;
+                }
```
vittorioromeo, line 306: Pull the `if` statement outside of the loop, so that we have `if (renderAbove) { for{...} } else { for{...} }`.

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -228,11 +266,19 @@ void HexagonGame::draw()
...
+                if(renderingAbove)
+                {
+                    wallQuads3DTop[k].position += newPos;
+                    wallQuads3DTop[k].color = overrideColor;
+                }
+                else
+                {
+                    wallQuads3D[k].position += newPos;
+                    wallQuads3D[k].color = overrideColor;
+                }
```
vittorioromeo, line 281: Pull the `if` statement outside of the loop, so that we have `if (renderAbove) { for{...} } else { for{...} }`.

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -212,11 +242,19 @@ void HexagonGame::draw()
...
+                if(renderingAbove)
+                {
+                    pivotQuads3DTop[k].position += newPos;
+                    pivotQuads3DTop[k].color = overrideColor;
+                }
+                else
+                {
+                    pivotQuads3D[k].position += newPos;
+                    pivotQuads3D[k].color = overrideColor;
+                }
```
vittorioromeo, line 257: Pull the `if` statement outside of the loop, so that we have `if (renderAbove) { for{...} } else { for{...} }`.

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -147,14 +150,27 @@ void HexagonGame::draw()
 
     if(Config::get3D())
     {
-        const float depth(styleData._3dDepth);
+        const float aboveMain(-styleData._3dLayerOffset - 1);
+        const bool renderAbove(aboveMain > 0 && !styleData._3dMainOnTop);
+        const float layerOffset(renderAbove ? -1.f : styleData._3dLayerOffset);
+        const float depth(styleData._3dDepth - renderAbove * aboveMain);
```
vittorioromeo, line 156: Some comments would be appreciated here, I don't find this code obvious, especially what `aboveMain` means and the role of `layerOffset`. Also maybe rename the variables a bit to be more meaningful, like `mustRender3DAboveMain` instead of `renderAbove`?

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -147,14 +150,27 @@ void HexagonGame::draw()
...
-        pivotQuads3D.reserve(numPivotQuads * depth);
-        playerTris3D.reserve(numPlayerTris * depth);
+        if(renderAbove)
+        {
+            wallQuads3DTop.reserve(numWallQuads * aboveMain);
+            pivotQuads3DTop.reserve(numPivotQuads * aboveMain);
+            playerTris3DTop.reserve(numPlayerTris * aboveMain);
+        }
+
+        if(depth > 0)
```
vittorioromeo, line 168: Would be nice to have a comment explaining why this only happens when `depth > 0`.

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -175,20 +191,34 @@ void HexagonGame::draw()
...
+
         const auto adjustAlpha = [&](sf::Color& c, const float i)
         {
             SSVOH_ASSERT(styleData._3dAlphaMult != 0.f);
 
             const float newAlpha =
                 (static_cast<float>(c.a) / styleData._3dAlphaMult) -
-                i * styleData._3dAlphaFalloff;
+                (styleData._3dAlphaMirror ? std::abs(i + 1) : i) *
+                    styleData._3dAlphaFalloff;
```
vittorioromeo, line 211: Ditto, this expression is pretty complicated. Either comments or maybe splitting it up would help.

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -175,20 +191,34 @@ void HexagonGame::draw()
...
 
             c.a = Utils::componentClamp(newAlpha);
         };
 
-        for(int j(0); j < static_cast<int>(depth); ++j)
+        for(int j(0);
+            j < static_cast<int>(renderAbove ? depth + aboveMain : depth); ++j)
```
vittorioromeo, line 217: Extract `static_cast<int>(renderAbove ? depth + aboveMain : depth)` to a `const` variable outside of the loop.

src/SSVOpenHexagon/Core/HGGraphics.cpp
```diff
@@ -175,20 +191,34 @@ void HexagonGame::draw()
...
         {
-            const float i(depth - j - 1);
+            const bool renderingAbove(j >= depth && renderAbove);
+            const float jAdj(j - depth * renderingAbove * (depth >= 0));
+            const float i(depth - j - 1 + layerOffset);
```
vittorioromeo, line 221: Ditto, especially confused about `jAdj`.

vittorioromeo (changes requested):
In favour of merging this feature, but I am having trouble understanding the changes a bit. I think some comments or better variable names or refactoring would help.

include/SSVOpenHexagon/Core/HexagonGame.hpp
Bauumm: True I tried implementing this in another way initially but ended up not using it and forgot to remove it from the header.

src/SSVOpenHexagon/Core/HGGraphics.cpp
Bauumm: yes

src/SSVOpenHexagon/Core/HGGraphics.cpp
Bauumm: Yeah, it's a bit confusing because i didn't want to split it into two loops, it's so it can switch to the next vertex container in the same loop. It's probably a better idea to just split it into two loops.

PR #379, opened 2022-09-09
Allow pack devs to specify font and font size for messages

Bauumm:
This is again a change for 1.92 parity so I can use the old font and size messages accordingly, so it doesn't look broken. Additionally custom fonts probably allow for some interesting new effects in levels too.


include/SSVOpenHexagon/Core/HexagonGame.hpp
```diff
@@ -205,6 +205,7 @@ class HexagonGame
     std::optional<SwapParticleSpawnInfo> swapParticlesSpawnInfo;
     float nextPBParticleSpawn{0.f};
     float pbTextGrowth{0.f};
+    float messageTextCharacterSize{32.f};
```
vittorioromeo, line 208: This likely needs to be reset between levels or even between resets of the same level. Maybe make it part of `HGStatus`? That should get reset automatically.

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -169,6 +169,21 @@ void HexagonGame::initLua_Utils()
         .doc(
             "Force-swaps (180 degrees) the player when invoked. If `$0` is "
             "`true`, the swap sound will be played.");
+
+    addLuaFn(lua, "u_setMessageFont", //
+        [this](std::string mFontId) {
```
vittorioromeo, line 174:
```suggestion
        [this](const std::string& mFontId) {
```

src/SSVOpenHexagon/Core/HGScripting.cpp
```diff
@@ -169,6 +169,21 @@ void HexagonGame::initLua_Utils()
...
+            messageText.setFont(
+                assets.getFontOrNullFont(getPackId() + "_" + mFontId));
```
vittorioromeo, line 176: I believe we can use the `concat` utility here to create the final string.

vittorioromeo (changes requested):
Hey, thanks for the PR! I am happy to add the feature. A few things:

1. See inline comments.

2. Testing! Have you tested this? Have you tested that the font and font size is correctly restored after reset or changing levels? Have you tested what happens when the user tries to load an invalid font or uses something like a negative font size? Please also add a test Lua level to this PR.

Bauumm:
I have tested loading invalid fonts, it just displays a dot in place of the message. Setting the font size to a negative value seems to make the message invisible.

Morxemplum:
As much as I love to see the enthusiasm for 1.92 compatibility, there is a good reason why Imagine (the 1.92 font) was ditched. It was done so for licensing purposes, as we had to change to another font if Vee wanted to sell the game on Steam. 

It was temporarily Forced Square, but due to the limited character set and the inability to modify that font directly, is what lead me to create the game's current font Open Square. I understand that some people like the look of Imagine, but unless someone is willing to buy the license, then you'll be running into the same copyright dilemma that pack developers already have to deal with when it comes to music.

Just my two cents on this.

Bauumm:
Well if I understand the issue correctly, the font could not be included with the game as it is commercial. However shouldn't it be possible to include the font in non-commercial level packs?

Morxemplum:
I am not a legal expert, so I can't answer that question. You got that first sentence right.

PR #381, opened 2022-09-09
Add onPreUnload core function

Bauumm:
I'm not sure if I should put all changes for parity in one big pull request rather than a lot of small ones now, but here is another one either way. This change is very specific to allow the function to run after timeline, levelStatus and a few other things have already been reset, while the lua environment has not. This way levels can actually access their lua variables when unloading, the old onUnload is kept for backwards compatability as well as the ability to tell if an attempt is a first attempt. This is also a very important change to make the level_change event that I implemented in my converter function properly since it needs to restart in the level it changed (There is a level that heavily relies on this mechanic.).

### DIFF
--- a/src/SSVOpenHexagon/Core/HexagonGame.cpp
+++ b/src/SSVOpenHexagon/Core/HexagonGame.cpp
@@ -700,6 +700,7 @@ void HexagonGame::newGame(const std::string& mPackId, const std::string& mId,
     inputImplCCW = inputImplCW = false;
     playerNowReadyToSwap = false;
 
+    runLuaFunctionIfExists<void>("onPreUnload");
     lua = Lua::LuaContext{};
     calledDeprecatedFunctions.clear();
     initLua();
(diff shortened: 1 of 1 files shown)


vittorioromeo (changes requested):
> I'm not sure if I should put all changes for parity in one big pull request rather than a lot of small ones now

Small and independent PRs are easier to review and merge than big ones, so please keep doing what you're doing :)

---

Regardless, could you explain a little bit more what the use case for `onPreUnload` is? I am not sure I see the benefit of adding this function.

Furthermore, I'd like to see some description of this function that can be added to the Lua documentation as part of this PR, could be just a comment here on GitHub.

Bauumm:
> Regardless, could you explain a little bit more what the use case for `onPreUnload` is? I am not sure I see the benefit of adding this function.
Well I created the function for multiple reasons:
1. It allows accessing the lua environment when the level is unloaded
2. It allows adding a timeline event to the next run of the level, this is very important to properly recreate the `level_change` event, since it should restart the level it changed to earlier (Rotational Adventures from Imaginative Leftovers heavily relies on this mechanic.)
3. It adds an onUnload that is called in the same place as the onUnload in 1.92 without sacrificing backwards compatability with new packs
> Furthermore, I'd like to see some description of this function that can be added to the Lua documentation as part of this PR, could be just a comment here on GitHub.
A core function that is called right before the lua environment is reset (before onUnload). This can be useful to influence the next attempt based on data from the last attempt.

vittorioromeo (approved):
LGTM. Will also update docs.

PR #383, opened 2022-09-17
add frametime parameter to onRenderStage

Bauumm:
This is another small one. It just gives the onRenderStage core function frametime as second parameter which allows for more accurately timed shader effects, since `l_getLevelTime()` only updates 240 times per second.


vittorioromeo:
Looks reasonable to me -- will merge as soon as I can test. Also self-reminder to update the documentation of this function.

vittorioromeo (approved):
LGTM.

PR #385, opened 2022-09-20
Make manual (Beat)Pulse control update immediately

Bauumm:
This one is mainly for a few levels in 1.92 that use these effects to an extreme extent, for those I need to be able to set the pulse values more than 240 times per second.


vittorioromeo:
Hmm, how does this help exactly? Aren't pulse and beatPulse values used 240 times per second anyway?

Bauumm:
Read again
> more than 240 times per second
This helps with some 1.92 levels that rely on some weird effects that happen when this is done (even on low refresh rate monitors).

Bauumm:
An example would be [Foreshock from Imaginative Leftovers](https://www.youtube.com/watch?v=4hPBary8yHE).

vittorioromeo (approved):
LGTM

PR #386, opened 2022-09-24
Add text shaders

Bauumm:
This pr adds the ability to modify the text color with fragment shaders. However it is a bit more unintuitive than the other RenderStages. So a shader that sets the outline to red and the text to green could look like this:
```glsl
#version 130
uniform vec4 color0;
uniform sampler2D font;


void main()
{
	vec4 color = gl_Color.rgb == color0.rgb ? vec4(1,0,0,1) : vec4(0,1,0,1);
	gl_FragColor = color * texture(font, gl_TexCoord[0].xy);
}
```
with `color0` being `s_getColor(0)` in order to set an outline color.

### DIFF
--- a/_RELEASE/Packs/experimental/Scripts/Levels/text_shader.lua
+++ b/_RELEASE/Packs/experimental/Scripts/Levels/text_shader.lua
@@ -1,8 +1,9 @@
+u_execDependencyScript("ohvrvanilla", "base", "vittorio romeo", "utils.lua")
+
+
 function onInit()
 	text_shader = shdr_getShaderId("text.frag")
-
-	-- RenderStage 8 is text
-	shdr_setActiveFragmentShader(8, text_shader)
+	shdr_setActiveFragmentShader(RenderStage.TEXT, text_shader)
 
 	-- make sure the text outline color isn't swapping
 	s_setMaxSwapTime(0)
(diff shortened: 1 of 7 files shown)


vittorioromeo:
Looks interesting! Could you add an example level using this as part of the PR, so that I can test it easily? Thank you.

_RELEASE/Packs/experimental/Scripts/Levels/text_shader.lua
```diff
@@ -0,0 +1,20 @@
+function onInit()
+	text_shader = shdr_getShaderId("text.frag")
+
+	-- RenderStage 8 is text
+	shdr_setActiveFragmentShader(8, text_shader)
```
vittorioromeo, line 5: There is an enum in `utils.lua` for this purpose, we should add the render stage there

```suggestion
u_execDependencyScript("ohvrvanilla", "base", "vittorio romeo", "utils.lua")

-- ...

	-- RenderStage 8 is text
	shdr_setActiveFragmentShader(RenderStage.TEXT, text_shader)
```

vittorioromeo (changes requested):
Just a minor thing. Also, I am currently travelling for work and unable to test the changes for a few more days, unfortunately.

vittorioromeo:
LGTM, thanks!

PR #390, opened 2022-10-08
Fix text appearing in wrong position for a tick

Bauumm:
This PR fixes the bug where a message appears at the wrong position at first and is positioned correctly in the next tick.
This bug is especially noticable in levels that create new messages very frequently.


vittorioromeo:
Thanks!

PR #392, opened 2022-10-15
Update music_timeline branch to latest changes

fentras:
Updates https://github.com/vittorioromeo/SSVOpenHexagon/pull/346 and https://github.com/vittorioromeo/SSVOpenHexagon/issues/335.

This is an updated version of [music_timeline](https://github.com/vittorioromeo/SSVOpenHexagon/tree/music_timeline) branch that has been merged with [develop](https://github.com/vittorioromeo/SSVOpenHexagon/tree/develop) branch (latest commit https://github.com/vittorioromeo/SSVOpenHexagon/commit/7f7d1665fd82dd92ead9c29109d2c3810271d601).
I've also added an example level to showcase the new functions.

While building and reviewing the pull request I have made some observations:

- When building you will get `__builtin_memcpy(__s1, __s2, __n)` warnings but they don't break the build process.
- If you set `a_evalTo` or (`a_evalToMs`) to zero seconds it will work only at the start of the level and will not repeat once the music loops back at the start. Having `a_evalToMs(10)` makes it work sometimes but not every time. Having anything above a second will work every time. I assume it is related to the global timeline which manages to skip evaluation (`a_evalTo` and `a_evalToMs`) methods with times close to zero at the start.
- `e_messageAdd*` seems to have a random delay from zero to 20 milliseconds before showing the message but I assume it's just tied to the screen refresh rate.
- If you put a singe song in the pack it will be used for every level even if you explicitly state to use a different one in a level's `.json`, but that doesn't seem to be related to `music_timeline` branch.

In general this is a very nice addition and will help introduce more music-focused levels.

https://user-images.githubusercontent.com/26721650/196008984-b58e49f7-2876-41df-a39f-be8bbade1078.mp4


fentras:
I completely forgot about https://github.com/vittorioromeo/SSVOpenHexagon/pull/346#issuecomment-852674353. Making it a draft for now.

Bauumm:
actually wait, look: https://github.com/vittorioromeo/SSVOpenHexagon/pull/346/commits/6a30c61a21c3154259b3267ec7133a36fbddf373 tho im not sure if this is a good way to do it, since it'd imo make more sense to just have the segment be randomly selected based on the seed instead of saving the segment itself.

fentras:
> have the segment be randomly selected based on the seed

That seems like a good idea! I'll update it soon.

fentras:
> That seems like a good idea! I'll update it soon.

But they never did.

PR #397, opened 2022-11-01
Allow overwriting all style colors

Bauumm:
This PR adds override colors for every color in a style which is only used if the alpha isn't 0. This way pack developers can specifically overwrite colors in a style which is a much requested feature. Additionally I also added the ability to set the color of single background panels and 3D layers.
I was also thinking about specifying if colors should be used in a different way, but settled for the alpha 0 check as the 3D override color already does this and it still allows transparent colors by having the styles color be transparent and overwriting the rest if really needed (it's an edge case in my opinion).
This PR is also useful for making 1.92 levels android compatible as shaders, which I use to reimplement the old style systems quirks, do not work there at the moment.


Bauumm:
Converting to draft as I realized I'll need to add some more stuff and it needs some fixing once #378 is merged.

PR #398, opened 2022-11-10
Fix wrong author in level info

Bauumm:
The level info during gameplay/replay shows the pack author instead of the level author. This PR makes it show the level author instead.


vittorioromeo:
Fixing manually to avoid conflicts. Thanks.

PR #399, opened 2022-11-13
fix dependency shaders

Bauumm:
For some reason the game looked for dependency shaders in the scripts folder, this should be fixed with this PR.


vittorioromeo:
Whoops, thanks!

PR #401, opened 2022-11-26
Fix replay path issues on windows

Bauumm:
Some testing in wine and on windows by other people has revealed that windows doesn't like certain characters in file paths. However some 1.92 packs use these in their name/id, so I need this to be fixed without changing that, which is what this PR does. It just replaces the invalid characters with spaces now.

### DIFF
--- a/include/SSVOpenHexagon/Core/HexagonGame.hpp
+++ b/include/SSVOpenHexagon/Core/HexagonGame.hpp
@@ -379,9 +379,12 @@ private:
     void sideChange(unsigned int mSideNumber);
 
     // Draw methods
-    void drawText_TimeAndStatus(const sf::Color& offsetColor, const sf::RenderStates& mStates);
-    void drawText_Message(const sf::Color& offsetColor, const sf::RenderStates& mStates);
-    void drawText_PersonalBest(const sf::Color& offsetColor, const sf::RenderStates& mStates);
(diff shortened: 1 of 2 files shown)


Bauumm:
Also related to this it seems that windows doesn't like long file paths. Which https://github.com/vittorioromeo/SSVOpenHexagon/pull/369 doesn't exactly improve. I did however find a way to somewhat fix this (it still has issues if the directory name becomes too big) by putting a manifest like this inside the exe file:
```xml
<?xml version='1.0' encoding='utf-8' standalone='yes'?>
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
  <application xmlns="urn:schemas-microsoft-com:asm.v3">
    <windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
      <ws2:longPathAware>true</ws2:longPathAware>
    </windowsSettings>
  </application>
</assembly>
```
This can be done using `mt.exe` from the windows sdk (this works in wine as well):
```
mt.exe -manifest this.manifest -outputresource:SSVOpenHexagon-Console.exe
```
but I don't know how that would be automated using cmake (the feature for that that exists seems to only be for projects using msvc).

vittorioromeo (approved):
LGTM, thanks!

PR #403, opened 2023-03-29
Make CI generated executables runnable

Bauumm:
This PR just makes some minor changes to the workflow (e.g. copying the correct libs or using the right directory) in order to create an actually functional executable.
While running it natively may cause some errors (such as `libFLAC.so.8` missing), it seems to always work with the steam runtime (test like this: `LD_PRELOAD=libstdc++.so.6 LD_LIBRARY_PATH=. ~/.steam/root/ubuntu12_32/steam-runtime/run.sh ./SSVOpenHexagon`, the environment variables will need to be put into the run script for a steam release of course (and both are necessary for it to work despite `libstdc++.so.6` not being shipped))


vittorioromeo:
@Bauumm: could you please fix the conflicts? Thanks for the PR, I will merge.

Bauumm:
Still has conflicts after rebasing? huh

Bauumm:
Just recreated it now, i think i messed something up here. https://github.com/vittorioromeo/SSVOpenHexagon/pull/406

PR #405, opened 2023-04-02
Random level/pack selection

Ena-Shepherd:
- Hitting LeftCtrl will select a random level on your current pack
- LeftCtrl + Focus selects a random pack
- Compatible with favorites


src/SSVOpenHexagon/Core/MenuGame.cpp
```diff
@@ -1015,6 +1014,49 @@ void MenuGame::initInput()
...
+
+            if (lvlDrawer->levelDataIds->size() == 1)
+                return;
+
+            randomLevel = ssvu::getRndI(0, lvlDrawer->levelDataIds->size());
+
+            if (randomLevel >= lvlDrawer->levelDataIds->size())
+                setIndex(0);
+            else
+                setIndex(randomLevel);
```
vittorioromeo, line 1030: Why are all these checks necessary? Wouldn't the following just always work, or am I missing something?

```suggestion
            setIndex(ssvu::getRndI(0, lvlDrawer->levelDataIds->size()));
```

Ditto for the other two similar functions.

vittorioromeo (changes requested):
Thanks for the PR. Not opposed to this feature, just a bit uncomfortable with the hardcoded keys.

src/SSVOpenHexagon/Core/MenuGame.cpp
Ena-Shepherd: Without the checks, line 1025 sometimes returns what seems to be undefined behavior like "-124852".
This creates an out of bounds error segfault and crashes the program.
Another exception to check is if the pack contains only 1 level, like tutorial pack, or the favorites list.

There is also 2 Keybinds intended to act as the combo keys, but i checked on both Shift Keys instead of Focus, so i will rework on this. (didn't find how to check on Focus directly).

vittorioromeo: `ssvu::getRndI(a, b)` returns a number between `a` and `b-1`. So `ssvu::getRndI(0, lvlDrawer->levelDataIds->size())` should always give a valid index.

The only case where it can go wrong is if `lvlDrawer->levelDataIds->size() == 0`, because it can underflow to a very large number. Maybe that's the only required check?

vindowvipre: Is there any chance you could add a bind for "random level in random pack" too? It is very commonly used in osu, for example.

Ena-Shepherd: > Is there any chance you could add a bind for "random level in random pack" too? It is very commonly used in osu, for example.

Yes, if Vee is ok with that

PR #406, opened 2023-04-05
 Make CI generated executables runnable #403

Bauumm:
This PR just makes some minor changes to the workflow (e.g. copying the correct libs or using the right directory) in order to create an actually functional executable. While running it natively may cause some errors (such as `libFLAC.so.8` missing), it seems to always work with the steam runtime (test like this: `LD_PRELOAD=libstdc++.so.6 LD_LIBRARY_PATH=. ~/.steam/root/ubuntu12_32/steam-runtime/run.sh ./SSVOpenHexagon`, the environment variables will need to be put into the run script for a steam release of course (and both are necessary for it to work despite `libstdc++.so.6` not being shipped))

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
-
+        
     - name: Install dependencies
       run: |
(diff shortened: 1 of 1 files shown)


Bauumm:
Closing as the this PR is no longer up to date (sfml is linked statically now)

PR #410, opened 2023-04-15
Add CI for mac osx

Bauumm:
This PR is based on #340 and adds a few more workarounds to make the most recent version of the game compile.
I'm still waiting for someone who owns a mac to see if the game actually works. But if it does we could have osx builds on steam as well hopefully.


Bauumm:
Alright, I got someone to test it now. So the game does actually work on OSX with these changes. (The only issue atm seems to be that the user needs to give permission to the game like 10 times for it to use the bundled frameworks)

M0REKZ:
Omg

Bauumm:
Does that play nicely with steam?

M0REKZ:
Oh, i forgot that this is a steam game lol, well, an app bundle wont cause problems because they would be folders inside a folder named "Open Hexagon.app"

However about permissions, it looks to be related to notarization, but since im not a steam dev, i couldnt know if its _really strictly_ necessary to notarize, Normally when i download an not-notarized app from a website i do a Right-click -> Open to bypass it the first time it launches.

esteluk:
I might try to poke at the app bundle a bit, but the `libsdkencryptedappticket.dylib` should be updated to at least version 1.53a in order to compile successfully on Apple Silicon machines (but 1.56 aligns with the current version used for other platforms).

Similarly the Discord SDK also needs an arm64 slice. I can't find an official version that combines the two into a single dylib, but it seems easy enough to combine them locally.

Bauumm:
Well I don't have a mac for testing (I relied entirely on actions for this) and I don't think actions have an apple silicon runner, so there's no way I can test it. But if you do manage to get something working feel free to open a PR (possibly even to my fork's branch so your changes would show up in this PR as well, but it doesn't really matter).

